PR #12, opened 2024-02-19
inputField: add dot color option

matteo4375:
Adds option to input-field to change the color of the dots when you type


matteo4375:
closing because pull request for this already exists #11

PR #17, opened 2024-02-20
Use local time for $TIME variable

CIAvash:
Before this change, system's time zone was not used to return the time


vaxerski (approved):
thats the problem with living in utc

PR #20, opened 2024-02-20
input_field: add font_color, font_size, font_family

fufexan:
Adds support for customizing color, size and font family of the input field text.


vaxerski:
you can set those via pango markup tho?

fufexan:
Then why is there `font_family` for the label?

vaxerski:
legacy reasons before it was supported, plus, fallback.

fufexan:
Alright, closing then.

PR #21, opened 2024-02-20
nix: fix hm module

jacekpoz:
changed cfg from config.services.hyprlock to config.programs.hyprlock

cc: @fufexan


fufexan:
Shit forgot about that. Ty.

PR #24, opened 2024-02-20
input-field: allow configure dots size

bvr-yr:
`dots_size` being scale of input-field height and clamped to `0.2 - 0.8`
`dots_spacing` being scale of dot's absolute value and clamped to `0.0 - 1.0`

maybe numbers are subject to change, but seems reasonable


vaxerski (approved):
lgtm wiki mr needed

fufexan:
What about dots' color?

PR #26, opened 2024-02-20
Add support for 90/270 degree rotated monitors

mauricekraus:
Currently there is an issue with rotated setups
image

Added support for monitors rotated at 90/270 degrees to fix display issues with rotated setups.

### DIFF
--- a/src/core/Output.cpp
+++ b/src/core/Output.cpp
@@ -11,7 +11,13 @@ static void handleGeometry(void* data, wl_output* output, int32_t x, int32_t y, 
 
 static void handleMode(void* data, wl_output* output, uint32_t flags, int32_t width, int32_t height, int32_t refresh) {
     const auto POUTPUT = (COutput*)data;
-    POUTPUT->size      = {width, height};
+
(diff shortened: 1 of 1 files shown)


src/core/Output.cpp
```diff
@@ -11,7 +11,13 @@ static void handleGeometry(void* data, wl_output* output, int32_t x, int32_t y,
 
 static void handleMode(void* data, wl_output* output, uint32_t flags, int32_t width, int32_t height, int32_t refresh) {
     const auto POUTPUT = (COutput*)data;
-    POUTPUT->size      = {width, height};
+
+    // handle portrait mode
+    if (POUTPUT->transform == WL_OUTPUT_TRANSFORM_270 || POUTPUT->transform == WL_OUTPUT_TRANSFORM_90) {
```
vaxerski, line 16: misses flipped transforms :P

vaxerski:
oh and no {} around short ifs

src/core/Output.cpp
mauricekraus: Sure sorry!

PR #28, opened 2024-02-20
config: add source parsing

### LINKED ISSUE #15: Sourcing file doesn't work.
```
source ~/.config/colors.conf
```
or 
```
source /home/username/colors.conf
```
Variables present in those files are not working in `hyprlock.conf`.
Is it meant to be this way?

And how you implemented different color in the label,(source preview image)?

mauricekraus:
This allows for out**source**ing configs into other files. Fixes #15

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -1,5 +1,21 @@
 #include "ConfigManager.hpp"
+#include "../helpers/MiscFunctions.hpp"
 #include <filesystem>
+#include <glob.h>
+#include <cstring>
+
+static Hyprlang::CParseResult handleSource(const char* c, const char* v) {
+    const std::string      VALUE   = v;
+    const std::string      COMMAND = c;
+
+    const auto             RESULT = g_pConfigManager->handleSource(COMMAND, VALUE);
+
+    Hyprlang::CParseResult result;
+    if (RESULT.has_value())
+        result.setError(RESULT.value().c_str());
+    return result;
+}
+
 
 static std::string getConfigDir() {
     static const char* xdgConfigHome = getenv("XDG_CONFIG_HOME");
@@ -15,7 +31,7 @@ static std::string getMainConfigPath() {
 }
 
 CConfigManager::CConfigManager() : m_config(getMainConfigPath().c_str(), Hyprlang::SConfigOptions{.throwAllErrors = true, .allowMissingConfig = true}) {
-    ;
+    configCurrentPath = getMainConfigPath();
 }
 
 void CConfigManager::init() {
@@ -52,6 +68,8 @@ void CConfigManager::init() {
     m_config.addSpecialConfigValue("label", "font_family", Hyprlang::STRING{"Sans"});
     m_config.addSpecialConfigValue("label", "halign", Hyprlang::STRING{"none"});
     m_config.addSpecialConfigValue("label", "valign", Hyprlang::STRING{"none"});
+
+    m_config.registerHandler(&::handleSource, "source", {false});
 
     m_config.commence();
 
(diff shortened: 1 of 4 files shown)


src/config/ConfigManager.cpp
```diff
@@ -132,3 +150,38 @@ std::vector<CConfigManager::SWidgetConfig> CConfigManager::getWidgetConfigs() {
...
+            if (std::filesystem::exists(value)) {
+                Debug::log(WARN, "source= skipping non-file {}", value);
+                continue;
+            }
+
+            Debug::log(ERR, "source= file doesnt exist");
+            return "source file " + value + " doesn't exist!";
+        }
+
+        m_config.parseFile(value.c_str());
```
vaxerski, line 182: you probably should update configCurrentPath no? so that `./lol.conf` called from inslde `./configs/pope.conf` can go to `./configs/lol.conf`

src/helpers/MiscFunctions.cpp
```diff
@@ -0,0 +1,25 @@
+#include "MiscFunctions.hpp"
+
+std::string absolutePath(const std::string& rawpath, const std::string& currentPath) {
+    auto value = rawpath;
+
+    if (value[0] == '~') {
+        static const char* const ENVHOME = getenv("HOME");
+        value.replace(0, 1, std::string(ENVHOME));
+    } else if (value[0] != '/') {
+        auto currentDir = currentPath.substr(0, currentPath.find_last_of('/'));
```
vaxerski, line 10: I hate this. Why not just `canonical( currentPath + rawpath )`

src/config/ConfigManager.cpp
mauricekraus: yep, missed that.

src/helpers/MiscFunctions.cpp
```diff
@@ -0,0 +1,20 @@
...
+    // Handling the case where rawpath starts with '~'
+    if (!rawpath.empty() && rawpath[0] == '~') {
+        static const char* const ENVHOME = getenv("HOME");
+        return std::filesystem::path(ENVHOME) / path.relative_path().string().substr(2);
+    }
+    // Handling e.g. ./, ../
+    else if (path.is_relative()) {
+        const std::filesystem::path currentDir = std::filesystem::path(rawcurrentpath).parent_path();
+
+        return std::filesystem::canonical(currentDir / path);
```
vaxerski, line 16: this can throw if the thing couldn't be found. We should return an empty string if tomething throws IMO or return an `std::optional` (I'd prefer `std::expected` but clang will shit itself)

mauricekraus: Okay, I've given my best, but I'm a bit confused if some of these error cases can even occur in handleSource. However, due to absolutePaths optionality, I needed to handle it anyway. I tested multiple configurations (invalid paths, etc.), and it should work now

src/helpers/MiscFunctions.cpp
```diff
@@ -0,0 +1,24 @@
...
+    // Handling e.g. ./, ../
+    else if (path.is_relative()) {
+        const std::filesystem::path currentDir = std::filesystem::path(rawcurrentpath).parent_path();
+
+        auto finalPath = currentDir / path;
+        if (exists(finalPath))
+            return std::filesystem::canonical(currentDir / path);
+        else
+            return {};
+    } else {
```
vaxerski, line 21: useless else. Just return. all if blocks always return.

vaxerski (commented):
rest lgtm, just please slap a clang-format

mauricekraus:
No it should be formatted correctly :sweat_smile:

vaxerski (approved):
thats better. Thanks!

PR #35, opened 2024-02-21
Add placeholder_color

luyu-wu:
Added placeholder_color configuration option that changes placeholder text color!
Hopefully everything is ok and I can become a more major contributer to Hypr stuff in the future (I got pretty moved by the blogpost).
Honestly the implementation is sketch af though, although it works on all cases I've tested it.

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -116,14 +116,15 @@ in {
       };
 
       placeholder_text = mkOption {
-        description = "The placeholder text color of the input field";
+        description = "The placeholder text of the input field";
         type = str;
         default = "<i>Input Password...</i>";
       };
-      font_color = mkOption {
-        description = "The font color of the input field";
+
+      placeholder_color = mkOption {
+        description = "The placeholder text color of the input field";
         type = str;
-        default = "rgb(120, 120, 120)";
+        default = "rgba(120, 120, 120,0.5)";
       };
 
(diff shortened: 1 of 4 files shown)


vaxerski:
placeholder text can already be changed with pango markup

matt1432:
> placeholder text can already be changed with pango markup

I feel like this would be a nice abstraction instead of having to do pango markup. I personally had to go through issues to figure it out. (that might be a skill issue on my part but still)

vaxerski:
it's clear in the docs even with a link?

plurpio:
this could be useful as "authentication failed" doesn't use pango markdown

vaxerski:
I'd prefer configuring the text there and not color

bvr-yr:
what about adding `fail_color` instead
and maybe on failure change both placeholder and outer color?

one problem with fail placeholder color being reverted, like it is now:
if i use `#1b1b1b` and `#e4e4e4` it's invisible
its ok grayish background, but not if fg/bg are nearly inverted by themselves

bvr-yr:
u mean just use markup like for normal placeholder?

vaxerski:
mhm

bvr-yr:
gonna look at it later
i suppose it is possible to extract parsed color to reuse?

vaxerski:
dont think so

PR #40, opened 2024-02-21
Nix: HM module: add dots options + convert input fields and labels to type = listOf submodule

leon-erd:
Convert the input fields and labels to type = listOf submodule to allow the creation of multiple labels / input fields
cc: @fufexan


fufexan (approved):
LGTM, thanks!

PR #46, opened 2024-02-21
Fallback to background color when background image not found

### LINKED ISSUE #45: Missing background image leaves empty lockscreen
I'm working on a utility that screenshots my monitors, messes up the images, and uses those for hyprlock's background images. I'm a pretty awful coder so frequently the images aren't where they should be when I test it.

If hyprlock cannot load the background image(s) specified in the config it draws nothing for the background leaving your workspaces visible for the world to see with the widgets drawn on top. 

I think it should probably fallback to the background color specified in the config or the default background color if even that is missing. If a user really wants a transparent lockscreen they can specify a color with a lower alpha value. 

I've written a hack around the problem and will submit a PR, this issue is just for reference.

DerekSauer:
Confirms if the background image was actually loaded and logs a warning if not.

If no asset exists for a background image, fall back to background color. 

Tested by having a background image where hyprlock expects it. Background is rendered properly. Deleted the background image from disk and hyprlock reverts to background color. Deleted background color in config and hyprlock reverts to default color. 

Fixes #45 

Disclaimer: I haven't written any C++ code since I was in college 20 years ago so please review this.

### DIFF
--- a/src/renderer/widgets/Background.cpp
+++ b/src/renderer/widgets/Background.cpp
@@ -6,17 +6,20 @@ CBackground::CBackground(const Vector2D& viewport_, const std::string& resourceI
 }
 
 bool CBackground::draw(const SRenderData& data) {
-    if (!asset)
-        asset = g_pRenderer->asyncResourceGatherer->getAssetByID(resourceID);
 
-    // If background image does not exist, for whatever reason, fallback to color
-    if (!asset) {
+    if (resourceID.empty()) {
         CBox   monbox = {0, 0, viewport.x, viewport.y};
         CColor col    = color;
         col.a *= data.opacity;
         g_pRenderer->renderRect(monbox, col, 0);
         return data.opacity < 1.0;
     }
+
+    if (!asset)
+        asset = g_pRenderer->asyncResourceGatherer->getAssetByID(resourceID);
+
+    if (!asset)
+        return false;
 
     CBox     texbox = {{}, asset->texture.m_vSize};
 
--- a/src/renderer/widgets/Background.cpp
+++ b/src/renderer/widgets/Background.cpp
@@ -6,20 +6,17 @@ CBackground::CBackground(const Vector2D& viewport_, const std::string& resourceI
 }
 
 bool CBackground::draw(const SRenderData& data) {
+    if (!asset)
+        asset = g_pRenderer->asyncResourceGatherer->getAssetByID(resourceID);
 
-    if (resourceID.empty()) {
+    // If background image does not exist, for whatever reason, fallback to color
+    if (!asset) {
         CBox   monbox = {0, 0, viewport.x, viewport.y};
         CColor col    = color;
         col.a *= data.opacity;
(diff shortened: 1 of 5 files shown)


vaxerski (approved):
lgtm

src/renderer/widgets/Background.cpp
```diff
@@ -6,21 +6,18 @@ CBackground::CBackground(const Vector2D& viewport_, const std::string& resourceI
 }
 
 bool CBackground::draw(const SRenderData& data) {
+    if (!asset)
```
vaxerski, line 9: actually, I'd check for the resourceID being empty here too. `getAssetByID` does a lookup and if the resourceID is empty why would we lookup

DerekSauer: Another approach would be to add an additional check in CRenderer::getOrCreateWidgetsFor() to see if the desired file exists and push an empty string ("") as the resource ID if it does not. That would allow checking for an empty resource ID in background render and skip the resource lookup, which is wasteful since the resource will never exist anyway.

I didn't do it that way since I wanted to error check the PNG load to guard not only against missing files but also corrupted files and OOM and such. Cairo has a bunch of error states it can return when loading an image.

DerekSauer:
Now it only checks if the image file exists. If it does not, it assigns an empty resource ID allowing immediate drop to background color fallback instead of attempting a pointless asset lookup. 

This won't protect against corrupt images or JPEGs pretending to be PNG files but why would someone do that for their lockscreen, they'd only be hitting themselves.

Cairo still tries to load the file that isn't there but I guess it doesn't actually care if the load succeeded or not and OpenGL doesn't seem care if it gets nothing for data and creates an empty texture. You're still paying for some empty buffers but fixing that can wait for a refactor of the resource system.

I typed my login password so many times while working on this I can no longer type it properly.

vaxerski:
> I typed my login password so many times while working on this I can no longer type it properly.

`general:grace = 1000` and any mouse movement will unlock xd

> This won't protect against corrupt images or JPEGs pretending to be PNG files but why would someone do that for their lockscreen, they'd only be hitting themselves.

What if, we generate the asset, but attach a flag `failed` or something to indicate it failed generating

DerekSauer:
Again give it a good look over if you could. I'm a Rust cultist and not having the borrow checker to slap me around when I'm playing dangerous with derefs makes me feel like I'm doing a Tom Cruise stunt with no harness.

vaxerski:
ye uhm oops I kinda uhh conflict hehe

DerekSauer:
lol, your last few commits invalidated my reason for doing this anyway and much more elegantly. blurry screenshots are a go

DerekSauer:
Recent changes has made this PR no longer necessary. Thank for all your work @vaxerski, I don't know where you find the energy.

DerekSauer:
Recent changes has made this PR no longer necessary. Thank for all your work @vaxerski, I don't know where you find the energy.

vaxerski:
me neither

PR #73, opened 2024-02-23
Unbreak build with libc++

jbeich:
Fixes #70. Tested with Clang/libc++ 17

- libc++ 18 has [atomic floats](https://github.com/llvm/llvm-project/commit/de7fbfeef598) but libc++ < 18 may still be common on distros
- libc++ 19 has [tzdb](https://github.com/llvm/llvm-project/commit/d332d88b919f) and [soon](https://github.com/llvm/llvm-project/pulls?q=is%3Apr+is%3Aopen+current_zone) will have `current_zone`

PR #75, opened 2024-02-23
fix libc++ compatibility v2

MrDuartePT:
Fixes https://github.com/hyprwm/hyprlock/issues/73. Tested with Clang/libc++ 17 and 18

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -3,6 +3,7 @@
 #include <filesystem>
 #include <glob.h>
 #include <cstring>
+#include <mutex>
 
 static Hyprlang::CParseResult handleSource(const char* c, const char* v) {
     const std::string      VALUE   = v;


src/config/ConfigManager.cpp
```diff
@@ -3,6 +3,9 @@
 #include <filesystem>
 #include <glob.h>
 #include <cstring>
+#if defined(_LIBCPP_VERSION)
```
vaxerski, line 6: why is this guarded?

MrDuartePT: I notice that it only required for libc++
Well my knowledge of c/c++ is a bit bad, that can be wrong. At least work on my side

MrDuartePT: Well if yoi think is wrong I can take out

MrDuartePT: Think about that I should use:
`#if defined(__clang__)` since on gcc it dosent produce a error without mutex

MrDuartePT:
Probably tomorrow I can't change anything until the end of the day. 
So I just commit the change I said above

src/config/ConfigManager.cpp
vaxerski: just include it without guards

MrDuartePT: Ok

MrDuartePT: Done

vaxerski (approved):
cool, thanks!

PR #82, opened 2024-02-25
Allow send SIGUSR1 to unlock

### LINKED ISSUE #16: Unlock command.
There should be an unlock command to avoid red screen of death.
What I mean to say is in `swaylock`
if I do `killall -q -s SIGUSR1 swaylock` it unlocks it without giving the red screen of death.
I can add that to idle's `unlock_cmd` and then `loginctl unlock-session` should unlock it.

Makes it easier to unlock the device if connected to kdeconnect.
would be nice if something like `hyprlock unlock` is introduced or `killall -q -s <SIGNAL> hyprlock`

singer233:
Fix #16 .As title said this allow user to send SIGUSR1 to hyprlock process to unlock it.  hopefully it went well.
and sorry for my bad English

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -43,12 +43,7 @@ inline const wl_seat_listener seatListener = {
 };
 
 // end wl_seat
-static void handleUnlockSignal(int sigNum){
-    if (sigNum==SIGUSR1){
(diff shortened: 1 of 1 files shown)


vaxerski (approved):

vaxerski:
thanks!

Issue #86, opened 2024-02-26
[Bug] Some key events are falsely treated as invalid, resulting in the wrong password being submitted occasionally.


n-ameless:
This bug is a result of [this commit](https://github.com/hyprwm/hyprlock/commit/4acf8981b34eaa6c3d9f823108daede4c6a75726)

This can be reproduced by repeating inputs such as, Shift + G, or any letter repeatedly.

I tried my best to explain it as clearly as I could.

vaxerski:
can't repro. Shift+G inputs G as many times as I spam it

EastArctica:
> can't repro. Shift+G inputs G as many times as I spam it

I can reproduce this. However, I don't think @n-ameless described how to reproduce it clearly. Spam both the shift key and the g repeatedly and your console will be littered with `Invalid key down event` messages. I also experience this as well when typing my password, usually notice it with the symbols.

n-ameless:
> > can't repro. Shift+G inputs G as many times as I spam it
> 
> I can reproduce this. However, I don't think @n-ameless described how to reproduce it clearly. Spam both the shift key and the g repeatedly and your console will be littered with `Invalid key down event` messages. I also experience this as well when typing my password, usually notice it with the symbols.

Yes exactly this. My english skills were failing me last night I was very tired lol. Spamming both Shift and the same key consecutively (G for example) triggers this.

vaxerski:
spamming is not a useful metric, can you find an exact combination to trigger this?

Timing should not matter.

EastArctica:
This issue is not fully resolved yet. Ignoring n-ameless's latest comment, what was patched was this:
1. Shift down
2. A down
3. Shift up
4. A up (`stray release event` will be logged)
5. Shift down
6. A down (`key already pressed` will be logged)
7. Shift up
8. A up (`stray release event` will be logged)
[Video](https://streamable.com/xl336j)

However, #0d5115a patched that. Now the issue can be seen with the following:
1. Shift down
2. 1 down
3. Shift up
4. A down
5. A up
6. 1 up (`stray release event` will be logged)
7. Shift down
8. 1 down (`key already pressed` will be logged)
9. 1 up
10. Shift up
[Video](https://streamable.com/rcxiu9)

EastArctica:
Why are we monitoring the key symbols instead of the keys themselves? If we were to use the `key` passed in the arguments it *should* fix it? 
[...]

PR #92, opened 2024-02-26
core: use raw key values for tracking keypresses

EastArctica:
Switches from using sym values to the raw key value which solves #86


vaxerski (approved):
fair enough I guess

PR #96, opened 2024-02-28
Support background paths relative to the home directory

### LINKED ISSUE #38: [REQ] expand `~` character in path
In config file expand `background:path` variable, if it has `~` expand it to `$HOME` environment variable. This is convenient for anyone sharing their config in multiple devices

PaideiaDilemma:
I used the existing `absolutePath` function with `rawcurrentpath` set to "" in order to expand "~" with ENVHOME.
The resource ids stay the same.

When the path is relative, absolutePath returns an empty option and it would use the original string.
That code path is kind of useless and maybe it would be better to just expand "~" outside of absolutePath.
Behaviour would be the same though.

Fixes #38

### DIFF
--- a/src/renderer/AsyncResourceGatherer.cpp
+++ b/src/renderer/AsyncResourceGatherer.cpp
@@ -5,6 +5,7 @@
 #include <pango/pangocairo.h>
 #include <algorithm>
 #include "../core/hyprlock.hpp"
+#include "../helpers/MiscFunctions.hpp"
 
 std::mutex cvmtx;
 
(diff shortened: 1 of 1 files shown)


src/renderer/AsyncResourceGatherer.cpp
```diff
@@ -130,10 +131,11 @@ void CAsyncResourceGatherer::gather() {
             if (path.empty())
                 continue;
 
-            std::string id = std::string{"background:"} + path;
+            std::string id            = std::string{"background:"} + path;
+            const auto  absolute_path = absolutePath(path, "").value_or(path);
```
vaxerski, line 135: const is CAPS and no_snake

vaxerski (approved):
lgtm

Issue #97, opened 2024-02-28
Cant see if i typed wrong the password


mooncoffee1:
When we type the password wrong, it would be cool to see a error like a red msg or maybe a moving animation that the password is incorect, rn its just stays like that when i typed the wrong pass, and idk what just happened, like, its frozen or what? then i need to type it again. Its not a big of a deal but it would be a nice feature

vaxerski:
it says it in the middle, tho

mooncoffee1:
strange, mine doesnt, maybe i didnt configure it correctly, it just stays and waits when i type wrong passwd

bvr-yr:
yeah thats exactly what i was talking about [here](https://github.com/hyprwm/hyprlock/pull/35#issuecomment-1976641019)

`#c8c8c8` and `#363646` are almost inverted colors
and u get `Authentication failed` in `#373737` on `#363646`

PR #98, opened 2024-02-29
Show autentication failReason without additional input

PaideiaDilemma:
This makes it so that
1. The renderer is triggered when the onPasswordCheckTimer callback is fired. Previously you had to give an additional input for the renderer to be triggered and the fail reason to show.
2. The updateDots function now sets dots.currentAmout to 0, when a failID is present and PASSLEN is 0. That leads to the draw function to return false and the failReason to be displayed until another input happens.

Addresses #97

**Previous behavior (I pressed a key at the end)**

https://github.com/hyprwm/hyprlock/assets/78690852/df047d4b-7ab6-4fa9-9abb-b1844841821f

**Behavior introduced here** 

https://github.com/hyprwm/hyprlock/assets/78690852/b900be89-fe30-449b-a60f-441770c8490f


vaxerski (approved):
alright, thanks!

Issue #74, opened 2024-02-23
weird crash on unlocking


Matthieu-Sauer:
I tried to simply unlock by entering my password, and for some reasons it simply crashed the app.
I just tried again with valgrind, and I found double free in pam's usage that might cause it?
[log.txt](https://github.com/hyprwm/hyprlock/files/14390190/log.txt)

Also, under valgrind, I simply could not unlock my session no matter what, I always had a pam authentification error (but this is probably caused by my gentoo install).

vaxerski:
no clue why this happens. it's as if pam itself frees the thing twice. I don't call a single free()

REALERvolker1:
This happens on arch for me too sometimes.

Matthieu-Sauer:
found what happens. getlogin() seems to return null for me, making pam do breakdance with memory.
I have every single occurence of getlogin() fail.
I would suggest to replace getlogin() call with overload one (custom one)
Using `getpwuid(getuid())` as a backup plan do work.
Do you want me to do this in a PR? it should not be that long

Matthieu-Sauer:
by replacing, I mean doing something like this:

```cpp
std::string_view utils::getLogin()
{
    auto login = getlogin();
    if (!login) {
        auto pwd = getpwuid(getuid());
        login = pwd;
    }
    if (!login)
        throw std::runtime_error("Could not get username. unlocking WILL fail.");
    return login;
}
```

vaxerski:
feel free to mr

PR #104, opened 2024-02-29
replacing getlogin C call with a wrap

Matthieu-Sauer:
this is a fix of issue #74 . the crash was caused by getlogin not returning anything on my system.

### DIFF
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -65,7 +65,7 @@ IWidget::SFormatResult IWidget::formatString(std::string in) {
         Debug::log(ERR, "Error in formatString, username null. Errno: ", errno);
 
     IWidget::SFormatResult result;
-    replaceAll(in, "$USER", username ? *username : std::string{""});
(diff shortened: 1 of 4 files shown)


vaxerski (commented):
lgtm but needs a clang-format

Matthieu-Sauer:
should be done

src/renderer/widgets/IWidget.cpp
```diff
@@ -58,13 +59,13 @@ static std::string getTime() {
...
 
-    char* username = getlogin();
+    auto username = getUserName();
 
     if (!username)
         Debug::log(ERR, "Error in formatString, username null. Errno: ", errno);
 
     IWidget::SFormatResult result;
-    replaceAll(in, "$USER", std::string{username ? username : ""});
+    replaceAll(in, "$USER", username ? *username : std::string{""});
```
vaxerski, line 68: `getUserName().value_or("")`?

Matthieu-Sauer: yeah fixed (i'm so dumb)

vaxerski:
conflict to be fixed, btw

PR #109, opened 2024-03-01
Fade out timeout

PaideiaDilemma:
Hi!

A bit of a follow up to #98

This handles the placeholder text in combination with fadeOnEmpty better, as it allows the user to specify a timeout for fading out the password field.

It now behaves the same weather an authentication failure happens or weather I empty the password input field (via ESC or Backspace).

This is how it looks with a default timeout of 1 second:

https://github.com/hyprwm/hyprlock/assets/78690852/69578892-cb34-480c-a7d0-215adaea0bf6

### DIFF
--- a/src/renderer/widgets/PasswordInputField.cpp
+++ b/src/renderer/widgets/PasswordInputField.cpp
@@ -12,6 +12,7 @@ CPasswordInputField::CPasswordInputField(const Vector2D& viewport_, const std::u
     dt_space                 = std::any_cast<Hyprlang::FLOAT>(props.at("dots_spacing"));
     dots.center              = std::any_cast<Hyprlang::INT>(props.at("dots_center"));
     fadeOnEmpty              = std::any_cast<Hyprlang::INT>(props.at("fade_on_empty"));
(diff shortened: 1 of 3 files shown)


src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -199,7 +200,7 @@ bool CPasswordInputField::draw(const SRenderData& data) {
             forceReload = true;
     }
 
-    return dots.currentAmount != PASSLEN || data.opacity < 1.0 || forceReload;
+    return dots.currentAmount != PASSLEN || fade.animated || forceReload;
```
PaideiaDilemma, line 203: `data.opacity < 1.0 ` did not really make sense to me. Did I miss something about that?

PaideiaDilemma:
This way it rerenders during the timeout though. Should this be implemented via a callback instead?

src/renderer/widgets/PasswordInputField.cpp
vaxerski: uhh I think the renderer will take care of that anyways I think

vaxerski (commented):
a callback would be nice.

Issue #4, opened 2024-02-19
Pam entry


niksingh710:
Hyprlock does not seem to create a pam entry file.
In compare to swaylock as it does with the contents 

was planning to test if it will work with fingerprint (using fprint-grosshack)

Also on other note would it be possible to have path acceptance of `~/`?

vaxerski:
dafoq is a pam entry file

> Also on other note would it be possible to have path acceptance of ~/?

for backgrounds? yeah planned alongside jpegs and webms

niksingh710:
> dafoq is a pam entry file
`/etc/pam.d/` (Pluggable Authentication Modules)
I also don't know in depth but in short they perform authentication.
So swaylock has one and if I specify some other `.so` file it can authenticate.
```
#
# PAM configuration file for the swaylock screen locker. By default, it includes
# the 'login' configuration file (see /etc/pam.d/login)
#

auth      sufficient pam_fprintd.so
auth include login
```
i am not a good explainer but here as I added pam_fprintd.so then it allows swaylock to get authenticated using fingerprint.

vaxerski:
I'll see, alr

bvr-yr:
btw its better to set
```pam
auth include system-auth
```

i don't remember exactly what's the reason, probably related to autologin or absence of stuff like sddm or greetd. swaylock doesn't work without changing it, while [waylock](https://github.com/ifreund/waylock) ships `system-auth` by default

kragebein:
My setup might have been fucked by IPA. But for some reason hyprlock wanted to authenticate as "su". And my "su" is 100% local and my normal setup is using system-auth. 

so had to add auth include system-auth into /etc/pam.d/su

Now it works.

alba4k:
@bvr-yr wait does that mean creating that file will work?

bvr-yr:
yes just as root
```sh
echo "auth include system-auth" > /etc/pam.d/hyprlock
```

niksingh710:
> yes just as root
> 
> ```shell
> echo "auth include system-auth" > /etc/pam.d/hyprlock
> ```

added this tried to also add fprintd entry with gross hack seems like hyrlock does not support blank password input so it failed to log me in using fingerprint. :(

bvr-yr:
i was talking about `login` vs `system-auth`, as former doesn't work in some cases
i don't use fingerprints so can't tell

alba4k:
[...]

PR #115, opened 2024-03-01
Adding a pam configuration file

alba4k:
This pull request would make hyprlock use it's own pam configuration file instead of the `su` one

This allows the user to include random pam services into it without touching stuff required by other programs.

By default, this just includes the `/etc/pam.d/login` configuration, which contains, by default:
```
#%PAM-1.0
auth       required     pam_securetty.so
auth       requisite    pam_nologin.so
auth       include      system-local-login
account    include      system-local-login
session    include      system-local-login
password   include      system-local-login
```

Basically just asks for a password.

Creating a custom pam configuration is exactly what https://github.com/hyprwm/hyprlock/issues/4 is asking for, and that issue is therefore fixed by this PR

On a sidenote, I'm not really sure about this line in `CMakeLists.txt`:
```cmake
# [...]
install(FILES "pam/hyprlock" DESTINATION "/etc/pam.d")
```
Is this the best way to do that? I'm not sure if using an absolute path is the best approach, but I didn't find anything better. Of course it can still be modified by using `DESTDIR`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,3 +80,5 @@ protocol("unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml" "linux-dmabuf-unst
 
 # Installation
 install(TARGETS hyprlock)
+
+install(FILES "pam/hyprlock" DESTINATION "/etc/pam.d")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,8 @@ include_directories(
   "protocols/"
 )
 
+include(GNUInstallDirs)
+
 # configure
 set(CMAKE_CXX_STANDARD 23)
 add_compile_options(-Wall -Wextra -Wno-unused-parameter -Wno-unused-value
@@ -81,4 +83,5 @@ protocol("unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml" "linux-dmabuf-unst
 # Installation
 install(TARGETS hyprlock)
 
-install(FILES "pam/hyprlock" DESTINATION "/etc/pam.d")
+install(FILES "pam/hyprlock" DESTINATION "${CMAKE_INSTALL_FULL_SYSCONFDIR}/pam.d")
+
--- a/pam/hyprlock
+++ b/pam/hyprlock
@@ -1,0 +1,5 @@
+# PAM configuration file for hyprlock
+# the 'login' configuration file (see /etc/pam.d/login)
+
+auth        include     login
+
--- a/src/core/Password.cpp
+++ b/src/core/Password.cpp
@@ -30,7 +30,7 @@ std::shared_ptr<CPassword::SVerificationResult> CPassword::verify(const std::str
         const pam_conv localConv = {conv, NULL};
         pam_handle_t*  handle    = NULL;
 
-        int            ret = pam_start("su", getlogin(), &localConv, &handle);
+        int            ret = pam_start("hyprlock", getlogin(), &localConv, &handle);
 
         if (ret != PAM_SUCCESS) {
             result->success    = false;


vaxerski:
all good just need @fufexan to fix nix nix fix pls fix nix nix fix

alba4k:
Was about to ping but wasn't 100% completely sure if it was just me being me

vaxerski:
also I think trying `su` if regular fails would be a nice touch too.

Don't lock the user out :(

alba4k:
`login` already includes a ton of stuff, that file not working would mean the user is already locked out of his system anyway (since `/bin/login` wouldn't work), ergo there would be bigger issues than hyprlock not unlocking.

This is the same approach swaylock and i3lock use.

alba4k:
On a sidenote, I managed to SIGSEGV hyprlock using my modified login file:
```
#%PAM-1.0
auth       sufficient   pam_unix.so try_first_pass likeauth nullok
auth       sufficient   pam_fprintd.so
auth       required     pam_securetty.so
auth       requisite    pam_nologin.so
auth       include      system-local-login
account    include      system-local-login
session    include      system-local-login
password   include      system-local-login
```

To do this I would press enter on an empty password, then the fingerprint auth will start. Once fprintd times out tho, hyprlock just terminates. Swaylock (yeah, sorry, that's the obvious comparison) will just restart the authentication. Should I open a different issue for that?

Again, this is not related to this PR in particular but moreso to using fprintd

CMakeLists.txt
```diff
@@ -80,3 +80,5 @@ protocol("unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml" "linux-dmabuf-unst
 
 # Installation
 install(TARGETS hyprlock)
+
+install(FILES "pam/hyprlock" DESTINATION "/etc/pam.d")
```
fufexan, line 84:
```suggestion
install(FILES "pam/hyprlock" DESTINATION "${CMAKE_INSTALL_SYSCONFDIR}/pam.d")
```

As per https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html.

alba4k:
@fufexan using that just makes it install to `/pam.d/hyprlock`

fufexan:
@alba4k what about `CMAKE_INSTALL_FULL_SYSCONFDIR`?
In any case we can include `-DCMAKE_INSTALL_SYSCONFDIR=/etc` in the readme build command, but it should already exist either way.

CMakeLists.txt
alba4k: @fufexan using that just makes it install to `/pam.d/hyprlock`
does cmake always need to be called with `-DCMAKE_INSTALL_SYSCONFDIR=/etc`?

alba4k:
Sorry for deleting, added it as a comment to the suggestion. `CMAKE_INSTALL_FULL_SYSCONFDIR` doesn't work either

CMakeLists.txt
fufexan: Oops my bad, forgot we have to also add `include(GNUInstallDirs)` before the`install(...)`.

alba4k: Yes that worked

fufexan (approved):
LGTM

vaxerski:
> login already includes a ton of stuff, that file not working would mean the user is already locked out of his system anyway (since /bin/login wouldn't work), ergo there would be bigger issues than hyprlock not unlocking.

> This is the same approach swaylock and i3lock use.

My logic was more like someone forgor :skull: to copy the file.

alba4k:
oh so in hyprlock, in case the file is not found. That makes sense.

PaideiaDilemma:
As a nix user I had to do "security.pam.services.hyprlock = {};" for this to work.
Maybe that should be mentioned in the hm module like [swaylocks hm module](https://github.com/nix-community/home-manager/blob/master/modules/programs/swaylock.nix) does.

alba4k:
@fufexan

vaxerski:
it's your fault for not fallbacking to su

alba4k:
> also I think trying `su` if regular fails would be a nice touch too.

@vaxerski looking into that right now. pam_start seems to return 0 regardless of whether the file exists or not, should I just manually check?

I'd just leave everything as-is tbh, as that is what every other program would usually do (again, take swaylock as an example).

vaxerski:
why not try regardless if any fail

fufexan:
> @fufexan

ye?

alba4k:
> ye?

https://github.com/hyprwm/hyprlock/pull/115#issuecomment-1974771483
Not familiar with nix stuff

alba4k:
> why not try regardless if any 

Because I'd have no idea how to implement that in the existing code, having never worked with pam myself :P

fufexan:
@alba4k done in 97548ec0ad318d3d10b2f52d2dccca58cec98bae.

alba4k:
ping @PaideiaDilemma not me ahah

vaxerski:
I swear to god amma revert this, I added the pam file and yet STILL can't login

vaxerski:
why did you make the pam file 600 ffs

vaxerski:
added su fallback in https://github.com/hyprwm/hyprlock/commit/f9fe60c7eb66c3a5e6ef56b66697de569a8479f4

Jackaed:
Can we make the su fallback optional @vaxerski? I'm trying to configure hyprlock to not try and use fprintd, despite the fact that my `su` does trigger fprintd auth, but even if I remove it from the hyprlock PAM config, if I put in an incorrect password, it still calls fprintd and then I have to wait for that to timeout before I can put a new attempt in. 

I'm not convinced that having `su` as a fallback by default is a good idea in general, if people need that functionality they can just include their `su` pam file within the hyprlock pam file. I'm yet to see any similar projects that behave similarly to hyprlock in this regard.

alba4k:
When you input a wrong password you get prompted for fingerprint but that likely has nothing to do with your su, rather with your hyprlock config

Jackaed:
I believed that to be the case as well, but fprint is not in my PAM config in `/etc/pam.d/hyprlock`, and removing fprint from my `su` pam configuration causes the system to no longer unlock with my fingerprint. I get this output from hyprlock when inputting an incorrect password:

```
[ERR] auth: Authentication failed for hyprlock
[ERR] auth: Authentication failed for su
```

Neither of these errors occur if I just input the correct password. I haven't looked at the code, but based on this, it seems like it uses `su` as a fallback if the unlock attempt wasn't successful.

PaideiaDilemma:
Guys the latest git version does not have a su fallback anymore. You can specify the pam module used by hyprlock in the config. Check the wiki :)

bvr-yr:
> `su` is still used as fallback btw

really? no fallback for me, I'm getting faillocks like intended.
What is not intended, faillock also happens with grace enebled. @PaideiaDilemma that's a bug or smth wrong on my end with pam?

PaideiaDilemma:
> `su` is still used as fallback btw

It is, but not at time of authentication, but at startup.
From the wiki:

> If the module isn’t found in `/etc/pam.d`, “su” will be used as a fallback

PR #119, opened 2024-03-02
Fix unlock sync

PaideiaDilemma:
e7fca0abd494fbc3c364ff380c7169522a281a14 fixes a red screen that is quite easily reproducible by moving the mouse a lot during the grace period.
I have yet to figure out how unlock is getting called multiple times consecutively, but now we print an error in this case and just continue, since we have already set m_bTerminate to true.

5b93762583be107bb6b0574eb619ee97195c4f67 fixes the hangs mentioned in #114.


vaxerski (approved):
good stuff

PR #124, opened 2024-03-02
Change install directory of hyprlock's pam configuration file

BranchManager:
I was having issues getting Hyprlock to work on Fedora. My password was never being authenticated correctly. The root of the problem seemed to be the Password.cpp file on line 34 where "su" was changed to "hyprlock" . However turns out the hyprlock pam configuration file was being installed in the /usr/local/ dir which is not where pam seems to look for conf files.

TLDR: So I changed where hyprlock's pam configuration file is being installed to "/etc/pam.d" in the CMakeLists.txt

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -9,11 +9,31 @@ self: {
   inherit (lib.modules) mkIf;
   inherit (lib.options) mkOption mkEnableOption;
 
-  boolToString = x: if x then "true" else "false";
+  boolToString = x:
+    if x
+    then "true"
+    else "false";
   cfg = config.programs.hyprlock;
 in {
   options.programs.hyprlock = {
-    enable = mkEnableOption "Hyprlock, Hyprland's GPU-accelerated lock screen utility";
+    enable =
+      mkEnableOption ""
+      // {
+        description = ''
+          Whether to enable Hyprlock, Hyprland's GPU-accelerated lock screen utility.
+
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
g

vaxerski:
I'm a baller, shot caller, I'm an arch linux installer

vaxerski:
@fufexan why is nix unhappy

fufexan:
Same exact reason as in the original PR. You can't access /etc in the nix build sandbox. And neither should we hardcode /etc anyway.

BranchManager:
Does it work on arch ? Haven't used it in a while.

vaxerski:
I guess you can `-DCMAKE_INSTALL_SYSCONFDIR:STRING="/etc/"`?

vaxerski:
thats why I am in favor of a `su` fallback

BranchManager:
I thought about changing it back to su but figured the code was changed for a reason. And I thought about adding that flag to read me for fedora users at least. So I can add that flag later tonight.

vaxerski:
fuffiee....

image

I think no distro sets this to etc. We _should_ hardcode it, and you can override it for nix.

BranchManager:
I was just gonna take your idea and potentially make an su fallback. If "hyprlock" fails try with the su parameter. Don't know if that is a good idea though

vaxerski:
I committed that, it works well

PR #125, opened 2024-03-03
hidden-input-field: add seed for random colors in hidden input field outline

p4rfait:
The colors in the outline and quadrant of the hidden input field aren't random because it lacks a seed; it always repeats the same color pattern: yellow, yellow, blue, blue, magenta, pink, green, magenta and so on. I'm unsure if this is the intended behavior, but this should fix it.


vaxerski (approved):
woopsie, thanks!

PR #126, opened 2024-03-03
Test if user is submiting empty string for password

BranchManager:
Fedora user back again. Not sure if many people were having this issue but if I don't type anything and hit enter hyprlock would crash and I would just get a red screen.  Added a check to see if the password is not an empty string before authenticating.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -703,9 +703,11 @@ void CHyprlock::onKey(uint32_t key, bool down) {
         if (m_sPasswordState.passBuffer.length() > 0)
             m_sPasswordState.passBuffer = m_sPasswordState.passBuffer.substr(0, m_sPasswordState.passBuffer.length() - 1);
     } else if (SYM == XKB_KEY_Return || SYM == XKB_KEY_KP_Enter) {
-        Debug::log(LOG, "Authenticating");
+        if (m_sPasswordState.passBuffer.length() > 0) {
+            Debug::log(LOG, "Authenticating");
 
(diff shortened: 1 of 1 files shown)


vaxerski:
thing is, aren't empty passwords allowed?

BranchManager:
There may need to be another solution then. I was not aware of that. I'm not sure if it is that case for anyone else but if I submit an empty password hyprlock crashes and I cannot continue my session.

vaxerski:
I guess since nobody should be idiotic enough to have an empty password I'll allow this.

vaxerski (approved):

BranchManager:
That's what I thought but I'll work on a fix, hopefully over the week, to allow an empty password.

PaideiaDilemma:
I think this breaks usage with pam_fprintd no?
Havn't used a fprintd in a while but I guess it theoretically would word with hyprlock.
When you use fprintd you specify an empty password to trigger the fingerprint reader.

vaxerski:
I don't use that so if anyone does and this breaks it for them, feel free to submit a mr

opticblu:
> I think this breaks usage with pam_fprintd no? Havn't used a fprintd in a while but I guess it theoretically would word with hyprlock. When you use fprintd you specify an empty password to trigger the fingerprint reader.

Yes, pam_fprintd auth needs "empty" password support for a screen locker, as nothing will be typed in the password field

matt1432:
Yeah I was using fprintd and now it doesn't work anymore.

I can still just type any character and then enter to unlock with fingerprint but it would be nice to allow empty passwords

PR #130, opened 2024-03-04
input-field: add dots rounding option

bvr-yr:
Follow-up 49ef1d306c92bbfef057272ae0bdefe8aea8d0c7

`-2` kinda follow input-field rounding

PR #127, opened 2024-03-03
Add rounding option to input field

matteo4375:
Adds a rounding option to input field so you're not forced to have a pill shaped input field

PR #132, opened 2024-03-05
Fix nix home manager module

matteo4375:
Fixes dots_rounding and rounding


fufexan (commented):
You also have to define the options, otherwise this won't work.

matteo4375:
the options are already defined in #127 and #130

fufexan (approved):

PR #142, opened 2024-03-06
Input fade out timeout #3

PaideiaDilemma:
Uhm I think I overthought the last try in #139 a bit.

Here the timer is added/canceled directly in CPasswordInputField::updateFade.


vaxerski (commented):
looks alright, just needs a wiki mr and a fix for the conflict

vaxerski (approved):
thanks!

PR #147, opened 2024-03-07
label: fix rendering with multiple timers that fire at the same time

### LINKED ISSUE #134: [BUG] Label is not updating without input-field
If input field is hidden - the label is not updating

My config

input-field {
    monitor =
    size = 200, 50
    outline_thickness = 3
    dots_size = 0.33 # Scale of input-field height, 0.2 - 0.8
    dots_spacing = 0.15 # Scale of dots' absolute size, 0.0 - 1.0
    dots_center = true
    # dots_rounding = -1 # -1 default circle, -2 follow input-field rounding
    # outer_color = rgb(151515)
    inner_color = rgb(200, 200, 200)
    font_color = rgb(10, 10, 10)
    fade_on_empty = true
    placeholder_text = Password # Text rendered in the input box when it's empty.
    hide_input = false
    # rounding = -1 # -1 means complete rounding (circle/oval)
    position = 0, -50
    halign = center
    valign = center
}
label {
    monitor =
    text = cmd[update:1000] echo "<b><big> $(date +"%H:%M:%S") </big></b>"
    color = rgba(200, 200, 200, 1.0)
    font_size = 64
    font_family = Noto Sans
    position = 0, 16
    halign = center
    valign = center
}

The clock stuck at the time of locking and changes seconds only when input is visible
system:
Arch
hyprlock installed with yay

PaideiaDilemma:
The problem is that `getAssetByID` might return a `nullptr` if `asyncLoopState.busy` is true.
In that case the callback of the `asyncResourceGatherer` does trigger rendering, but the asset is not getting updated.

Not sure if you want to re=render like this or handle this case (`asyncLoopState.busy` beeing true) specifically.

Fixes #134

### DIFF
--- a/src/renderer/widgets/Label.cpp
+++ b/src/renderer/widgets/Label.cpp
@@ -114,7 +114,7 @@ bool CLabel::draw(const SRenderData& data) {
     shadow.draw(data);
     g_pRenderer->renderTexture(box, asset->texture, data.opacity);
 
-    return false;
+    return !pendingResourceID.empty();
 }
 
 void CLabel::renderSuper() {


vaxerski (approved):
lgtm

PR #148, opened 2024-03-07
label: higher precision resource ids

PaideiaDilemma:
When a label gets updated more than once a second, `time(nullptr)` does not have enough precision.


vaxerski (approved):
lgtm thanks bae

PR #150, opened 2024-03-08
core: terminate the poll thread

### LINKED ISSUE #149: [BUG] The process remains for a long time after unlocking
`hyprlock` doesn't return for a while, even after unlocking (5 seconds?)

this issue ha pretty much always been there.

PaideiaDilemma:
Poll will return when it catches any signal. 
I used `SIGUSR2` to register a signal that does nothing, so that we can use it for terminating the poll thread without any coredumps. 
To clarify the need for the signal handler: When we `pthread_kill` and poll is currently blocking, then it would not coredump anyways. But if we `pthread_kill` and the poll thread is currently not blocking, then it would core dump without a handler.

If you would signal SIGUSR2 to the whole process while hyprlock is running, then nothing would happen.

The `sigaction` change is just because it is more consistent and could prevent a future headache on a weird system.

Fixes #149

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -292,11 +292,11 @@ void CHyprlock::onGlobalRemoved(void* data, struct wl_registry* registry, uint32
 
 // end wl_registry
 
-static void registerSignalAction(int sig, void (*handler)(int)) {
+static void registerSignalAction(int sig, void (*handler)(int), int sa_flags = 0) {
     struct sigaction sa;
     sa.sa_handler = handler;
     sigemptyset(&sa.sa_mask);
-    sa.sa_flags = 0;
+    sa.sa_flags = sa_flags;
     sigaction(sig, &sa, NULL);
 }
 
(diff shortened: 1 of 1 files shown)


src/core/hyprlock.cpp
```diff
@@ -355,7 +366,7 @@ void CHyprlock::run() {
 
     std::thread pollThr([this, &pollfds]() {
         while (!m_bTerminate) {
-            int ret = poll(pollfds, 1, 5000 /* 5 seconds, reasonable. It's because we might need to terminate */);
+            int ret = poll(pollfds, 1, -1);
```
vaxerski, line 369: I don't like this. I'd keep it at 5s just in case.

vaxerski (approved):
yeah lgtm

PaideiaDilemma:
I thought about just using `SIGTERM`, since terminating hyprlock does not make any sense and will cause a red screen anyways.
Don't know if that would be considered bad though.

Otherwise we could use a real time signal, since they can also be used for application-defined purposes according to the man page.

vaxerski:
SIGRTMIN ?

PR #152, opened 2024-03-08
Fix label update

### LINKED ISSUE #105: Labels not getting updated on all screens
I have two 4K displays and universal hyprlock config meaning things should get applied to all outputs. The left screen only gets updated when something is typed in the password box, right one keeps up with the updates as expected.

<details>
  <summary>hyprlock.conf</summary>
  

  ```
general {
    disable_loading_bar = false
    grace = 5
}

background {
    monitor =
    path = screenshot
    color = rgba(0, 0, 0, 0)
    blur_passes = 3
    blur_size = 7
    noise = 0.0117
    contrast = 0.8916
    brightness = 0.8172
    vibrancy = 10
    vibrancy_darkness = 1.0
}

label {
    monitor =
    text = cmd[update:60000] echo "$(date +'%H:%M')"
    color = rgba(229, 123, 48, 1.0)
    font_size = 256
    font_family = ProggyClean Nerd Font, bold

    position = 0, 200
    halign = center
    valign = center
}

label {
    monitor =
    text = cmd[update:2000] echo "$(if [ $(cat /sys/class/leds/input*::capslock/brightness | head -n 1) = 1 ]; then echo "!!CAPS ON NERD!!";fi)"
    color = rgba(255, 0, 0, 1.0)
    font_size = 256
    font_family = ProggyClean Nerd Font, bold

    position = 0, 0
    halign = center
    valign = center
}

input-field {
    monitor =
    size = 200, 50
    outline_thickness = 3
    outer_color = rgba(15, 15, 15, 1)
    inner_color = rgba(200, 200, 200, 1)
    font_color = rgba(100, 0, 255, 1)
    fade_on_empty = true
    placeholder_text = "<i>Password...</i>"
}
  ```
  
</details>

hyprlock_fail_render

edit: also noticed that red screen gets triggered often when cursor is moved on the left screen during grace period, nothing gets posted in the logs though

PaideiaDilemma:
#147 Did not fully do the trick.

When  `frameCallback` is set, the `render` function will instantly return. That will make onAssetCallback not update the asset via renderSuper and the label get's stuck with it's current asset until the next render.

31cbb83b824605b5aef0516f0d113644a93fb131 is what fixes the problem.

ed9c1c29aafc76be3e9fd75b9697d66a7f239ec4 was just to check my sanity but it seemed like a sensible change so I left it in

The fix from #147 is not really needed any more since it is now handled via the callback.

Fixes #105

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -316,7 +316,7 @@ std::vector<std::unique_ptr<IWidget>>* CRenderer::getOrCreateWidgetsFor(const CS
             } else if (c.type == "input-field") {
                 widgets[surf].emplace_back(std::make_unique<CPasswordInputField>(surf->size, c.values));
             } else if (c.type == "label") {
-                widgets[surf].emplace_back(std::make_unique<CLabel>(surf->size, c.values, /* evil */ const_cast<CSessionLockSurface*>(surf)));
+                widgets[surf].emplace_back(std::make_unique<CLabel>(surf->size, c.values, surf->output->stringPort));
             }
         }
     }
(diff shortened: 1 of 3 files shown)


vaxerski (approved):
lgtm

PR #157, opened 2024-03-09
Handle slow pam verification

abihf:
For example when using systemd-homed with fido auth

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -371,6 +371,10 @@ void CHyprlock::run() {
             int ret = poll(pollfds, 1, 5000 /* 5 seconds, reasonable. Just in case we need to terminate and the signal fails */);
 
             if (ret < 0) {
+                if (errno == EINTR) {
+                    continue;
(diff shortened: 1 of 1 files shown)


src/core/hyprlock.cpp
```diff
@@ -371,6 +371,10 @@ void CHyprlock::run() {
             int ret = poll(pollfds, 1, 5000 /* 5 seconds, reasonable. Just in case we need to terminate and the signal fails */);
 
             if (ret < 0) {
+                if (errno == EINTR) {
```
vaxerski, line 374: style nit: no {} around short ifs

PaideiaDilemma:
`ret < 0` really shouldn't happen when you authenticate.

Could you check which signal gets sent with something like `strace -e 'trace=!all' ./build/hyprlock`?

PaideiaDilemma:
Ahh ok thanks. I guess they come from systemd-logind and are probably fine.

vaxerski (approved):

PR #158, opened 2024-03-09
background: handle missing assets

### LINKED ISSUE #155: Using non PNG image leaves empty passthrough background with no color
I use a script that chooses the wallpaper on each boot. Many of them are non-png. The problem is that I use the same image for the hyprlock background. When the image is non png, it will show the lock screen with all the open apps with no color background. This is a major problem with security.

PaideiaDilemma:
When there is only a background widget on the surface and the first render fails to get the resource from the gatherer the background will not get rendered.

490fe4bb42615c8f3684300b1dfbe8561781b44b fixes that

The other two commits allow the background to fall back to the background color if cairo fails to handle the path.
Hope TEXTURE_INVALID is the right thing to use for that.

Fixes #155


vaxerski (approved):
lgtm

PR #161, opened 2024-03-10
README: improve docs for dependencies

arijit79:
Improved the dependencies section in the README. Also removed `gles3.2` dependency as it is usually provided by mesa

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,16 +17,24 @@ Hyprland's simple, yet multi-threaded and GPU-accelerated screen locking utility
 ## Building
 
 ### Deps
+Hyprlock requires `hyprlang` >= 0.4.0. See the 
+[README](https://github.com/hyprwm/hyprlang/blob/main/README.md#building-and-installation) for 
+build instructions.
+
+You also need the following dependencies
  - wayland-client
  - wayland-protocols
+ - mesa
+ And the development libraries for the following:-
  - cairo
- - gles3.2
  - libdrm
  - pango
- - hyprlang>=0.4.0
  - xkbcommon
  - pam
- - mesa
+Development libraries are usually suffixed with `-devel` or `-dev` in most distro repos.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,16 +17,24 @@ Hyprland's simple, yet multi-threaded and GPU-accelerated screen locking utility
 ## Building
 
 ### Deps
+Hyprlock requires `hyprlang` >= 0.4.0. See the 
+[README](https://github.com/hyprwm/hyprlang/blob/main/README.md#building-and-installation) for 
+build instructions.
```
vaxerski, line 22: why this part? Most package managers are now on >= 0.4.0

arijit79: Yeah I think the build from source part wasn't actually required as it is already shipped by default in most distros repos. Maybe we can rephrase it to say that if its not shipped on any distro's repos, people can still jump to its build instructions. I will put it in my next commit.

vaxerski: that's default for any dep, I think the comment is redundant.

arijit79: Ok. I changed it in my latest commit.

README.md
```diff
@@ -17,20 +17,17 @@ Hyprland's simple, yet multi-threaded and GPU-accelerated screen locking utility
...
- And the development libraries for the following:-
- - cairo
- - libdrm
- - pango
- - xkbcommon
- - pam
+- wayland-client
+- wayland-protocols
+- mesa
+And the development libraries for the following:-
```
vaxerski, line 24: why is there a - at the end here xd

vaxerski (commented):
rest lgtm

README.md
arijit79: I just like to add a - in these types of lines, just have a habit of it. I will remove that if you don't like that.

vaxerski: ye please dont

vaxerski (approved):

PR #163, opened 2024-03-10
misc: path handling improvements

PaideiaDilemma:
Using weakly_canonical avoids a runtime error when the path does not exist.
`absolutePath` now does not check if the file exists. That is done later on anyways.


vaxerski (approved):
ok

PR #164, opened 2024-03-10
input-field: display utf-8 codepoint length

### LINKED ISSUE #129: [BUG] Each cyrillic symbol is displayed as 2 dots

PaideiaDilemma:
Only counts codepoints, which is fine, since you can only enter one at a time anyways.

Fixes #129


vaxerski (approved):
lgtm

PR #169, opened 2024-03-10
core: force update labels on `SIGUSR2`

bvr-yr:
reasons behind: https://github.com/hyprwm/hyprlock/pull/150#issuecomment-1987201385

by default force update is disabled and can be enabled with:
`[update:12345:1]` or `[update:12345:true]`

are you ok with parsing like that? i think it makes more sense than config option

### DIFF
--- a/src/core/Timer.hpp
+++ b/src/core/Timer.hpp
@@ -5,10 +5,11 @@
 
 class CTimer {
   public:
-    CTimer(std::chrono::system_clock::duration timeout, std::function<void(std::shared_ptr<CTimer> self, void* data)> cb_, void* data_);
+    CTimer(std::chrono::system_clock::duration timeout, std::function<void(std::shared_ptr<CTimer> self, void* data)> cb_, void* data_, bool force);
 
     void  cancel();
     bool  passed();
+    bool  canForceUpdate();
 
     float leftMs();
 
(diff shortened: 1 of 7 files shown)


vaxerski:
looks fine by me I think?

bvr-yr:
wait i'll allow `update:0:true`

src/core/Timer.hpp
```diff
@@ -19,5 +20,6 @@ class CTimer {
     std::function<void(std::shared_ptr<CTimer> self, void* data)> cb;
     void*                                                         data = nullptr;
     std::chrono::system_clock::time_point                         expires;
-    bool                                                          wasCancelled = false;
-};
\ No newline at end of file
+    bool                                                          wasCancelled     = false;
+    bool                                                          allowForceUpdate = false;
```
bvr-yr, line 24: or should I rename here? not to be the same as in [IWidget.hpp](https://github.com/hyprwm/hyprlock/pull/169/files#diff-6cf15676b59d5669af3c5248d657b158e129f110267c03965273f6d713b1a917R21)?

vaxerski: it fine

vaxerski (approved):
shoot to kill

PR #172, opened 2024-03-11
input-field: `check_color` + some fixes

bvr-yr:
`check_color` to change outer when waiting for check

- shadow animated on fade (not so heavy as i thought)
- resize extra width on fail kinda animated a bit now
- fixed fail counter not updated on empty password input (if no `fade_on_empty`)

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -261,6 +261,12 @@ in {
               default = -1;
             };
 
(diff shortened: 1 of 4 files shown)


bvr-yr:
damn! i've been fixing #174 now cause its my fault. nwm i'll submit in few minutes

PR #182, opened 2024-03-12
input-field: add capslock and numlock options

bvr-yr:
- `capslock_color`
- `numlock_color`
- `invert_numlock` change color when numlock off (for those who rely on numpad, like me)
- `bothlock_color` when both conditions

by default `color` options are `-1`, which means ignore state
that's ok? or maybe add default colors but leave ability to set `-1`
i don't know how to handle this correctly in Home Manager, as not on Nix
should i just use
```nix
capslock_color = ${toString background.capslock_color}
```
and which type to set then?

now function looks way more cleanly and other stuff can be added easier

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -90,68 +90,67 @@ in {
     backgrounds = mkOption {
       description = "Background configurations";
       type = listOf (submodule {
-        options =
-          {
-            monitor = mkOption {
-              description = "The monitor to apply the given wallpaper to";
-              type = str;
-              default = "";
-            };
-
-            path = mkOption {
-              description = "The path to the wallpaper";
-              type = str;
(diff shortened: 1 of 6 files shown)


bvr-yr:
can you test a bit on multi-monitor? there been some issues with colors not being updated properly to the end on one of monitors, i think i fixed that but still kinda laggy a bit rarely with > 300ms transition, perhaps just my gpu sucks here

vaxerski (approved):
lgtm

vaxerski:
for nix @fufexan

fufexan:
@bvr-yr added the nix options. If you feel the descriptions don't match well, feel free to change them.

bvr-yr:
nah that's good, thanks!

vaxerski (approved):
g

PR #186, opened 2024-03-13
label: add `$FAIL` and `$ATTEMPTS`

### LINKED ISSUE #179: [REQ] Allow drawing fail_text in another location
I'm using `input-field:hide_input = true`, and have a bunch of labels inside of the input-field circle. For this use-case, it's really weird to have the fail_text get drawn in the circle, under those labels.

Ideally, I'd love to be able to specify an offset for the fail_text from the input-field widget, allowing me to for example draw it below, rather than in the circle.

There is an important note here though, currently, fail_text will also expand the input-field to fit that fail_text inside of it. If the fail_text is rendered outside of the input-field, that obviously doesn't make sense. I suppose this could be solved with another variable like `expand_on_fail`, and if enabled, the expanding would need to support expanding input-field enough to fit the fail_text in another location too. I don't really care too much how this will be done, this is just one suggestion here.

### LINKED ISSUE #180: [REQ] Allow persistent fail_text
I currently `input-field:fail_text` primarily to show me the amount of failed attempts. This is useful because I can immediately see if someone has attempted to unlock my device while I was away.

However currently, once you start typing a new password again, the `fail_text` disappears, and the `outer_color` is reset back too from the `fail_color`, going back to the original state.

That means that someone could very easily hide that they attempted to login by just starting to type a new password, and then deleting it.

That means the only way for me to actually find out if there was a failed attempt made is to try and use a wrong password intentionally, to get the `fail_text` to appear again. That's obviously not good UX.

bvr-yr:
should resolve #179 
should resolve #180

`$ATTEMPTS[blabla]` can be used to show blabla or nothing when no failed attempts

https://github.com/hyprwm/hyprlock/assets/130279855/f68dc058-b7f1-448d-a1cb-c7196bfe8f96

one thing to mention, when `$ATTEMPTS` used in cmd, sometimes failasset in input-field doesn't draw
maybe smth could be done?
when `sleep 0.1` is the first command in `cmd`, all ok

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -307,14 +307,18 @@ static void handleUnlockSignal(int sig) {
     }
 }
 
+static void forceUpdateTimers() {
+    for (auto& t : g_pHyprlock->getTimers()) {
+        if (t->canForceUpdate()) {
+            t->call(t);
+            t->cancel();
+        }
+    }
+}
+
 static void handleForceUpdateSignal(int sig) {
     if (sig == SIGUSR2) {
-        for (auto& t : g_pHyprlock->getTimers()) {
-            if (t->canForceUpdate()) {
-                t->call(t);
-                t->cancel();
-            }
-        }
+        forceUpdateTimers();
     }
 }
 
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
lgtm

vaxerski:
needs a wiki mr

bvr-yr:
sure just thought you maybe against `[]` format

PR #194, opened 2024-03-15
core: fade out improvements

PaideiaDilemma:
Wanted the fade out animation for grace unlocks as well.
But also:
 - disables input once the fade has started
 - renders after the fade out started (needed for grace unlock, fine for normal unlocks as well I guess)
 - don't disable the fade out for other compositors (not really needed)

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -531,11 +531,9 @@ void CHyprlock::run() {
 }
 
 void CHyprlock::unlock() {
-    static auto* const PNOFADEOUT     = (Hyprlang::INT* const*)g_pConfigManager->getValuePtr("general:no_fade_out");
-    const auto         CURRENTDESKTOP = getenv("XDG_CURRENT_DESKTOP");
-    const auto         SZCURRENTD     = std::string{CURRENTDESKTOP ? CURRENTDESKTOP : ""};
+    static auto* const PNOFADEOUT = (Hyprlang::INT* const*)g_pConfigManager->getValuePtr("general:no_fade_out");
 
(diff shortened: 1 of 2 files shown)


vaxerski:
> don't disable the fade out for other compositors (not really needed)

Needed. Other compositors render a solid background behind it, and the fade out is idiotic in that case.

PaideiaDilemma:
yeah I just though it is fine because fade in kinda behaves the same, it does not look too bad (checked on sway) and you can turn it off anyways.

But I will undo it :)

src/core/hyprlock.cpp
```diff
@@ -530,6 +530,24 @@ void CHyprlock::run() {
     Debug::log(LOG, "Reached the end, exiting");
 }
 
+void CHyprlock::unlock() {
```
vaxerski, line 533: naming gets confusing here, `unlockChecked` and `unlockUnchecked`, maybe? Or something?

PaideiaDilemma: Agree. How about renaming `lockSession` and `unlockSession` to smth like `acquireSessionLock` and `releaseSessionLock`?

vaxerski: sure

vaxerski (approved):
lgtm

bvr-yr:
i'm also working on some fixes and have a question here
do we need `success_color`? or just fading back to outer is enough
because currently stuck `check_color` looks confusing when unlocking
probably just make it as an option and default to -1 = outer?

vaxerski:
idk

PR #198, opened 2024-03-15
input-field: various fixes

bvr-yr:
- correctly erase utf8 chars on backspace
- don't draw outer at all if `outline_thickness = 0`
- don't draw hidden input quadrants on fades. even if respecting `fade.a` and `data.opacity`, it's colored and still visible like quadrant (ugly)
- don't draw dots on unlock fadeout (they are flashing now)
- don't stuck on `check_color` on unlock fadeout
- now correctly checks `dots.currentAmount` against utf8 in draw return

I also made input-field fades to be a bit longer

### DIFF
--- a/src/renderer/widgets/PasswordInputField.cpp
+++ b/src/renderer/widgets/PasswordInputField.cpp
@@ -89,22 +89,20 @@ void CPasswordInputField::onFadeOutTimer() {
 }
 
 void CPasswordInputField::updateFade() {
-    const auto PASSLEN = g_pHyprlock->getPasswordBufferLen();
-
     if (!fadeOnEmpty) {
         fade.a = 1.0;
         return;
     }
 
-    if (PASSLEN > 0 && fade.allowFadeOut)
+    if (passwordLength > 0 && fade.allowFadeOut)
         fade.allowFadeOut = false;
 
-    if (PASSLEN > 0 && fade.fadeOutTimer.get()) {
+    if (passwordLength > 0 && fade.fadeOutTimer.get()) {
         fade.fadeOutTimer->cancel();
         fade.fadeOutTimer.reset();
     }
 
-    if (PASSLEN == 0 && fade.a != 0.0 && (!fade.animated || fade.appearing)) {
+    if (passwordLength == 0 && fade.a != 0.0 && (!fade.animated || fade.appearing)) {
         if (fade.allowFadeOut || fadeTimeoutMs == 0) {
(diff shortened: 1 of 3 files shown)


bvr-yr:
after c06afbe don't return on up events, instead just handle only down, but keep rendering

if keys pressed repeatedly, dots animation was kinda choppy
especially when pressing same key twice fast enough
I have two same characters at the end and very often last dot didn't render, but only after few seconds it did

haven't noticed any performance difference even on my shitty gpu

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -124,9 +122,9 @@ void CPasswordInputField::updateFade() {
 
     if (fade.animated) {
         if (fade.appearing)
-            fade.a = std::clamp(std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - fade.start).count() / 100000.0, 0.0, 1.0);
+            fade.a = std::clamp(std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - fade.start).count() / 150000.0, 0.0, 1.0);
         else
-            fade.a = std::clamp(1.0 - std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - fade.start).count() / 100000.0, 0.0, 1.0);
+            fade.a = std::clamp(1.0 - std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - fade.start).count() / 150000.0, 0.0, 1.0);
```
vaxerski, line 127: why this change?

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -175,6 +171,9 @@ bool CPasswordInputField::draw(const SRenderData& data) {
 
     bool forceReload = false;
 
+    passwordLength = g_pHyprlock->getPasswordBufferDisplayLen();
```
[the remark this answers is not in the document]
vaxerski: its fine

src/renderer/widgets/PasswordInputField.cpp
bvr-yr: seemed to quick, just to line-up with other animations. change it back?

vaxerski: yea

vaxerski (approved):

PR #199, opened 2024-03-17
Add missing destructors causing segfaults

stazio:
Hello!
I was running into a couple of segfaults caused because of the lack of de-initialization of certain components. I thought I would track it down and send in a fix!

The segfault I was really concerned by was caused by the lack of call to `eglTerminate` as it caused a segfault to get thrown from deep within `libnvidia-egl-wayland.so`.


vaxerski (approved):
makes sense, thanks!

Issue #170, opened 2024-03-10
pam_conversation not being handled properly


Mikilio:
I have been wishing for this program for a while now and was happy to discover it. It doesn't work on my setup where I use u2f authentication to login, which is supported by PAM, so I read through the code to see what is up, and I noticed that the pam_conversation part is not really handled well. Apparently the current solution is to spam PAM with the password until it unlock instead of having a proper conversation.

Actually, there already exists a default function called misc_conv to use as pam_conversation. To use it, we just need to specify STDIN_FILENO and STDOUT_FILENO in the thread and send it application data through a pipe.

This however requires to also change the way we display the conversation. Because the input field may not be used exclusively for the password anymore.

To figure this out, I already procrastinated a lot to read about PAM, so I'll be glad to make use of this knowledge in a PR. However, I don't really know modern C++ but rather pure C, so I may need assistance to add the "++" to my code.
It would also be nice to receive some guidance on how to utilize the input field. And how to echo messages from PAM if wanted.

vaxerski:
feel free to make a MR I review those anyways

PaideiaDilemma:
I naively implemented conv function by reading through the pam documentation as well and by checking how other applications do it.

Currently when the message type is `PAM_PROMPT_XXXX`, we just give pam the password. The message types that are not implemented are `PAM_RADIO_TYPE` and `PAM_BINARY_PROMPT`. Of course the pam configuration could prompt us for something other than the main password (e.g. 2fa). In that case we would need to keep the `handle` and get some more input from the user. That is the hard part. But piping the user input into a filedescriptor to use it with misc_conv is probably not the best approach as it is not that complicated to handle it in our own conv. (See https://github.com/linux-pam/linux-pam/blob/master/libpam_misc/misc_conv.c#L379)

But I think it shouldn't be too hard to use the placeholder text as the prompt message and get more input from the use via the existing input method.

Mikilio:
Firstly, thanks for being on board.

When doing authentications with PAM it can occur that there will be multiple prompts and answers. This is especially true with 2FA. Since the whole `pam_conversation` needs to occur on the single call of `pam_start` this means we need to implement a way for the `pam_conversation` to block and wait for user interaction. All this is already perfectly implemented in misc_conv, and it is also already included in the project.

Could you help me understand what advantage there is to implement our own `pam_conversation`?

[...]

PR #200, opened 2024-03-17
authclient supporting full PAM

Mikilio:
An authentication client that allows communicating with PAM in all possible ways via a Unix socket.
Solves #170.

### DIFF
--- a/src/core/Password.cpp
+++ b/src/core/Password.cpp
@@ -1,128 +1,133 @@
 #include "Password.hpp"
-#include "hyprlock.hpp"
 #include "../helpers/Log.hpp"
+#include "hyprlock.hpp"
 
 #include <security/_pam_types.h>
-#include <unistd.h>
-#include <sys/socket.h>
 #include <security/pam_appl.h>
+#include <sys/socket.h>
+#include <unistd.h>
 #if __has_include(<security/pam_misc.h>)
 #include <security/pam_misc.h>
 #endif
 
 #include <cstring>
+#include <expected>
 #include <thread>
 
-using namespace std;
-
 class AuthClient {
-  public:
-    int sock;
-    AuthClient(int sockfd) {
-        sock = sockfd;
-    };
-    //TODO: replace with std::expected errorhandling when clangd supports C++23
-    string question(const struct pam_message* msg) {
-        char data[10240];
+public:
+  int sock;
+  AuthClient(int sockfd) { sock = sockfd; };
+  std::optional<std::string> question(const struct pam_message *msg) {
+    char data[10240];
 
-        //TODO: maybe use protobufs or extra class?
-        int          mlen = strlen(msg->msg);
-        vector<char> packet(strlen(msg->msg) + sizeof(int));
-        const char*  style = static_cast<const char*>(static_cast<const void*>(&(msg->msg_style)));
-        copy(style, style + sizeof(int), back_inserter(packet));
-        copy(msg->msg, msg->msg + mlen, back_inserter(packet));
+    // TODO: maybe use protobufs or extra class?
+    int mlen = strlen(msg->msg);
+    std::vector<char> packet(strlen(msg->msg) + sizeof(int));
+    const char *style =
+        static_cast<const char *>(static_cast<const void *>(&(msg->msg_style)));
+    copy(style, style + sizeof(int), back_inserter(packet));
+    copy(msg->msg, msg->msg + mlen, back_inserter(packet));
 
-        if (send(this->sock, packet.data(), mlen + sizeof(int), 0) < 0) {
-            //ERROR:theres an error here
-            return string();
-        }
-        if (recv(this->sock, data, 10240, 0) < 0) {
-            //ERROR:there's an error here
-            return string();
-        }
-        //TODO: should I clear data?
-
-        //the response ist always just a string
-        return string(data);
+    if (send(this->sock, packet.data(), mlen + sizeof(int), 0) < 0) {
+      Debug::log(ERR, "auth: failed at sending question due to: \n{}\n",
+                 std::strerror(errno));
+      return {};
     }
+    if (recv(this->sock, data, 10240, 0) < 0) {
+      Debug::log(ERR, "auth: failed at retrieving response due to: \n{}\n",
+                 std::strerror(errno));
+      return {};
+    }
+    // TODO: should I clear data? overwrite with zeros
+
+    // the response ist always just a string
+    return std::string(data);
+  }
 };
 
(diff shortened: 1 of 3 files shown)


src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
-    const char*          pass      = static_cast<const char*>(appdata_ptr);
-    struct pam_response* pam_reply = static_cast<struct pam_response*>(calloc(num_msg, sizeof(struct pam_response)));
-
-    for (int i = 0; i < num_msg; ++i) {
-        switch (msg[i]->msg_style) {
-            case PAM_PROMPT_ECHO_OFF:
-            case PAM_PROMPT_ECHO_ON: pam_reply[i].resp = strdup(pass); break;
-            case PAM_ERROR_MSG: Debug::log(ERR, "PAM: {}", msg[i]->msg); break;
-            case PAM_TEXT_INFO: Debug::log(LOG, "PAM: {}", msg[i]->msg); break;
+using namespace std;
```
vaxerski, line 16: please fucking dont

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+
+class AuthClient {
+  public:
+    int sock;
+    AuthClient(int sockfd) {
+        sock = sockfd;
+    };
+    //TODO: replace with std::expected errorhandling when clangd supports C++23
+    string question(const struct pam_message* msg) {
```
vaxerski, line 25: can use `std::optional` I guess in the meantime

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+
+        //TODO: maybe use protobufs or extra class?
+        int          mlen = strlen(msg->msg);
+        vector<char> packet(strlen(msg->msg) + sizeof(int));
+        const char*  style = static_cast<const char*>(static_cast<const void*>(&(msg->msg_style)));
+        copy(style, style + sizeof(int), back_inserter(packet));
+        copy(msg->msg, msg->msg + mlen, back_inserter(packet));
+
+        if (send(this->sock, packet.data(), mlen + sizeof(int), 0) < 0) {
+            //ERROR:theres an error here
```
vaxerski, line 36: Debug::log it

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+        return string(data);
+    }
+};
+
+int conv(int num_msg, const struct pam_message** msgs, struct pam_response** response, void* appdata_ptr) {
+    int                          count = 0;
+    string                       reply;
+    vector<struct pam_response*> replies(num_msg);
+
+    AuthClient*                  authcl = (AuthClient*)appdata_ptr;
```
vaxerski, line 55: can be const auto

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+                 * *resp is a struct pam_response array and not an array of pointers.
+                 */
+                free(r->resp);
+                free(r);
+            }
+            return PAM_CONV_ERR;
         }
     }
-    *resp = pam_reply;
+    *response = *replies.data();
```
vaxerski, line 79: this is cursed

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+            for (struct pam_response* r : replies) {
+                //TODO: safe erasure?
+
+                /*
+                 * NOTE:pam_conv(3) - It is the caller's responsibility to release both,
+                 * this array and the responses themselves, using free(3). Note,
+                 * *resp is a struct pam_response array and not an array of pointers.
+                 */
+                free(r->resp);
+                free(r);
```
PaideiaDilemma, line 74: This means you don't have to free it since the caller in this case is the pam module not us xD

src/core/Password.cpp
```diff
@@ -11,20 +13,70 @@
...
+    AuthClient*                  authcl = (AuthClient*)appdata_ptr;
+
+    if (num_msg <= 0)
+        return PAM_CONV_ERR;
+
+    for (count = 0; count < num_msg; ++count) {
+        if (!(reply = authcl->question(msgs[count])).empty()) {
+            struct pam_response r = {reply.data(), 0};
+            replies[count]        = &r;
+        } else {
```
PaideiaDilemma, line 64: We need to check the pam msg type here and only call question for `PAM_PROMPT`.

PaideiaDilemma:
So since this is only rough start, would you mind me just taking a stab at implementing it @Mikilio?

src/core/Password.cpp
Mikilio: ok, I won't

src/core/Password.cpp
Mikilio: ok, I will

src/core/Password.cpp
Mikilio: yes, sir

src/core/Password.cpp
Mikilio: will do

src/core/Password.cpp
Mikilio: linux-util/login does this, though. I guess it could also be handled by front-end.

src/core/Password.cpp
Mikilio: I know. linux-util/login does that so `response` only ever points to a valid response array or whatever it pointed to before the function call. I just imitated that.

Mikilio:
Definitely, I actually don't have that much time, so I was planning to dedicate only some time each weekend. But if you're willing to wait, I'd also be happy to use this as a C++ learning opportunity, and I'll only bother you once a weekend with a review

src/core/Password.cpp
vaxerski: I meant `*replies.data()`

Mikilio: Should I do this?

```
*response = replies[0];
```

src/core/Password.cpp
```diff
@@ -1,76 +1,133 @@
 #include "Password.hpp"
-#include "hyprlock.hpp"
 #include "../helpers/Log.hpp"
+#include "hyprlock.hpp"
 
-#include <unistd.h>
+#include <security/_pam_types.h>
```
jbeich, line 5: Breaks build on FreeBSD which uses OpenPAM (also used on NetBSD, DragonFly, macOS):
```c++
src/core/Password.cpp:5:10: fatal error: 'security/_pam_types.h' file not found
    5 | #include <security/_pam_types.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
```

Try using `__has_include()` like a few lines below:
```suggestion
#if __has_include(<security/_pam_types.h>)
#include <security/_pam_types.h>
#endif
```

PR #201, opened 2024-03-17
input-field: new color features

bvr-yr:
- new `swap_font` option (see on videos)
- now when `outline_thickness = 0` (aka borderless) inner box color will be changed instead

this is how it looks

borderless with `swap_font`:
will change dots color to inner, that's for cases bright font looks ugly on colored box

https://github.com/hyprwm/hyprlock/assets/130279855/19d8b855-b1db-48e6-bcc0-ec33c30f458b

normal with `swap_font`:
will swap inner and font on check and failure

https://github.com/hyprwm/hyprlock/assets/130279855/c3db7b0f-e3d6-47d8-93ca-4bec19f44834

now i think we should deprecate `fail_transition` in favour of `color_transition`
how it is handled correctly?

### DIFF
--- a/src/renderer/widgets/PasswordInputField.cpp
+++ b/src/renderer/widgets/PasswordInputField.cpp
@@ -172,6 +172,10 @@ bool CPasswordInputField::draw(const SRenderData& data) {
 
     bool forceReload = false;
 
+    if (passwordLength == 0 && g_pHyprlock->getPasswordFailedAttempts() > failedAttempts)
+        forceReload = true;
+
+    failedAttempts = g_pHyprlock->getPasswordFailedAttempts();
     passwordLength = g_pHyprlock->getPasswordBufferDisplayLen();
     checkWaiting   = g_pHyprlock->passwordCheckWaiting();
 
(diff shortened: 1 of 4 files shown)


bvr-yr:
for some reason failasset stopped showing after certain commit when hitting enter on empty input, i added workaround for now

vaxerski:
why is this called swap_font?

bvr-yr:
i'm bad at namings. dunno, `swap_font_color`?

vaxerski:
more like that, yea, it deals with colors after all

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -171,13 +172,17 @@ bool CPasswordInputField::draw(const SRenderData& data) {
 
     bool forceReload = false;
 
+    if (passwordLength == 0 && g_pHyprlock->getPasswordFailedAttempts() > failedAttempts)
```
bvr-yr, line 175: I found culprit, it was initially working wrong since I allowed updating faillasset on empty imput. The color function was broken and incorrectly spamming `animated = true` causing redraw. Now it works correct but nothing else triggers `return true` in this function since dots = passlen on empty input. Color function is being called but only once and in checking state, thus not triggering `animated` because colors are equal.

i guess this is harmless

vaxerski (approved):
lgtm

src/renderer/widgets/PasswordInputField.cpp
vaxerski: hwat

bvr-yr: nvm, all ok

bvr-yr:
check please if everything is alright in last commit

vaxerski (approved):
idk looks reasonable

PR #202, opened 2024-03-17
config: add -c, --config commandline options

bvr-yr:
just copy-pasted from hypridle
strange it's not here already

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -30,7 +30,7 @@ static std::string getMainConfigPath() {
     return getConfigDir() + "/hypr/hyprlock.conf";
 }
 
-CConfigManager::CConfigManager(std::string configPath) : m_config(configPath.empty() ? getMainConfigPath().c_str() : configPath.c_str(), Hyprlang::SConfigOptions{.throwAllErrors = true, .allowMissingConfig = true}) {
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -24,6 +26,9 @@ int main(int argc, char** argv, char** envp) {
         else if (arg == "--quiet" || arg == "-q")
             Debug::quiet = true;
 
+        else if (arg == "--config" || arg == "-c")
```
vaxerski, line 29: missing check for argc

bvr-yr: done

src/config/ConfigManager.cpp
```diff
@@ -30,8 +30,8 @@ static std::string getMainConfigPath() {
     return getConfigDir() + "/hypr/hyprlock.conf";
 }
 
-CConfigManager::CConfigManager() : m_config(getMainConfigPath().c_str(), Hyprlang::SConfigOptions{.throwAllErrors = true, .allowMissingConfig = true}) {
-    configCurrentPath = getMainConfigPath();
+CConfigManager::CConfigManager(std::string configPath) : m_config(configPath.empty() ? getMainConfigPath().c_str() : configPath.c_str(), Hyprlang::SConfigOptions{.throwAllErrors = true, .allowMissingConfig = true}) {
```
vaxerski, line 33: you shouldn't allow a missing config in the case of a custom config path

bvr-yr: done

vaxerski (approved):
lgtm

Issue #203, opened 2024-03-18
Uses deprecated method to obtain username (doesn't work in some cases)


kcirick:
I tried using hyprlock but couldn't unlock (authentication error). After looking at the log file, I see the entries:

`[ERR] Error in formatString, username null. Errno:` 

Invesigating further, this is where the issue lies:
https://github.com/hyprwm/hyprlock/blob/2ae79757d5e5c48de2f4284992a6bfa265853a2d/src/renderer/widgets/IWidget.cpp#L85

Indeed, getlogin() on my machine returns (null). I found the following which seems to indicate this is a deprecated way to obtain a login name:
https://stackoverflow.com/a/4785195

Using `getpwuid(getuid())` I confirm it returns the correct username. Here is the swaylock implementation:
https://github.com/swaywm/swaylock/blob/f9ce3f193bc2f035790372b550547686075f4923/pam.c#L74

I think hyprlock should use this implementation?

kcirick:
Attaching a patch that makes it work for my system in hyprlock release version 0.2.0:

[hyprlock-0.2.0-username_fix-1.patch](https://github.com/hyprwm/hyprlock/files/14636920/hyprlock-0.2.0-username_fix-1.patch)

vaxerski:
was supposed to be fixed in #104 but that got abandoned. Feel free to make a MR superseding that one

kcirick:
Looked at that pull, but IMO there's no need to do anything fancy like that. I was also unsure what the outstanding conflict was on that pull.

Forgive me for being a n00b at git, I'll have to learn how to do PR/MR. Do I make a new Pull request?

vaxerski:
make a branch, commit your diff there, come back to this repo, click pull requests, new, and the rest is intuitive

PartyWumpus:
#204 has been merged, this issue should probably be closed?

PR #204, opened 2024-03-18
Use getpwuid(getuid()) instead of getlogin()

kcirick:
This is to fix getlogin() returning (null) under certain systems (e.g. https://github.com/hyprwm/hyprlock/issues/203 and https://github.com/hyprwm/hyprlock/pull/104)

### DIFF
--- a/src/core/Password.cpp
+++ b/src/core/Password.cpp
@@ -39,9 +39,9 @@ std::shared_ptr<CPassword::SVerificationResult> CPassword::verify(const std::str
 
     std::thread([this, result, pass]() {
         auto auth = [&](std::string auth) -> bool {
-            const pam_conv localConv = {conv, (void*)pass.c_str()};
-            pam_handle_t*  handle    = NULL;
-            auto uidPassword = getpwuid(getuid());
+            const pam_conv localConv   = {conv, (void*)pass.c_str()};
+            pam_handle_t*  handle      = NULL;
+            auto           uidPassword = getpwuid(getuid());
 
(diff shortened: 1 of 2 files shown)


src/core/Password.cpp
```diff
@@ -40,8 +41,9 @@ std::shared_ptr<CPassword::SVerificationResult> CPassword::verify(const std::str
         auto auth = [&](std::string auth) -> bool {
             const pam_conv localConv = {conv, (void*)pass.c_str()};
             pam_handle_t*  handle    = NULL;
+            struct passwd *passwd = getpwuid(getuid());
```
vaxerski, line 44: I'd rather drop the `struct` (this is C++) and just name the variable something like uidPassword

kcirick: I replaced `struct` with `auto` keyword with the suggested variable name.

src/renderer/widgets/IWidget.cpp
```diff
@@ -82,7 +83,8 @@ static std::string getTime() {
 
 IWidget::SFormatResult IWidget::formatString(std::string in) {
 
-    char* username = getlogin();
+    auto uidPassword = getpwuid(getuid());
+    char* username = uidPassword->pw_name;;
```
vaxerski, line 87: double ; here on accident I believe

kcirick: Thanks for your attention to detail! It's fixed now

vaxerski (commented):
lgtm, just needs a clang-format

vaxerski (approved):

PR #205, opened 2024-03-18
auth: implement a full pam conversation

### LINKED ISSUE #170: pam_conversation not being handled properly

PaideiaDilemma:
Hi!

Currently, on paper, we don't handle the pam conversation fully like it is intended by pam. (#107)
This is an implementation where we actually would handle it properly. At least I think so from what I gathered.

BUT, this kinda has diminishing returns, since you would only need this if you use a weird custom pam config or 2FA and I don't know if there are people who want a 2FA lockscreen.
I think fingerprint and yubikeys should work fine with the current solution (pls correct me if I am wrong).

Also having a fallback pam module is not really feasible this way. (To be fair personally I am not a fan of the su fallback, since it makes me have to wait longer when I fat finger my password)

What it gives you is the pam prompt and the ability to input multiple things in order to authenticate with pam.

**Implementation Notes**
Thought about using a socket like @Mikilio tried in #200.
But it seemed a bit overkill and I don't really want to send raw passwords over a socket if I don't have to.
This just uses a CV to wait for input.

I put this out to discuss. Still needs double checking. Also sorry for the single fat commit.

**Question**
I kinda think it would be cool to support a full conversation like this. But Idk if anyone would use it. Do we want it or no?

And sorry to @Mikilio for just overtaking this. I hope you don't mind :heart: 

Closes #170


vaxerski:
some rebasing fails, I sense

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -309,41 +305,55 @@ bool CPasswordInputField::draw(const SRenderData& data) {
...
+        return;
+
+    const auto AUTHFEEDBACK = AUTHFEEDBACKOPT.value();
+    if (AUTHFEEDBACK.text.empty() || placeholder.lastAuthFeedback == AUTHFEEDBACK.text)
+        return;
+
+    placeholder.isFailText       = AUTHFEEDBACK.isFail;
+    placeholder.lastAuthFeedback = AUTHFEEDBACK.text;
+
+    placeholder.asset = nullptr;
```
PaideiaDilemma, line 331: TODO: I guess we want to keep the old asset around, so that we always have one to render

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -176,18 +179,18 @@ bool CPasswordInputField::draw(const SRenderData& data) {
 
     failedAttempts = g_pHyprlock->getPasswordFailedAttempts();
```
bvr-yr, line 180: I've added this and check above to force draw on empty input, because nothing was triggering it and failasset was missing.
Here is [fixup.patch](https://github.com/hyprwm/hyprlock/files/14652880/fixup.patch) to use it directly in placeholder function, because now failasset is missing when typing directly after failure (without showing normal asset) and fail again, and also on empty input.

It also unbreaks some color changing.

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -423,7 +433,7 @@ void CPasswordInputField::updateColors() {
     col.stateNum  = col.invertNum ? !g_pHyprlock->m_bNumLock : g_pHyprlock->m_bNumLock;
     col.stateCaps = g_pHyprlock->m_bCapsLock;
 
-    if (placeholder.failID.empty()) {
+    if (!placeholder.isFailText) {
```
bvr-yr, line 436: Just keep this checks (from patch above) in mind, maybe there is a way to make them not so ugly xD after work is done.

src/renderer/widgets/PasswordInputField.cpp
PaideiaDilemma: Thanks for the patch! :)

src/core/Auth.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    m_sConversationState.waitingForPamAuth = false;
+
+    if (ret != PAM_SUCCESS) {
+        m_sConversationState.success    = false;
+        m_sConversationState.failReason = ret == PAM_AUTH_ERR ? "Authentication failed" : "pam_authenticate failed";
+        Debug::log(ERR, "auth: {} for {}", m_sConversationState.failReason, pam_module);
+        return false;
+    }
+
```
Mikilio, line 93: I don't mind at all. I feel like this taught me a lot actually since this implementation is a lot better, and truthfully I've also been thinking that a socket was overkill after implementing it. On that note: I think it might be relevant to call [pam_acct_mgmt](https://man7.org/linux/man-pages/man3/pam_acct_mgmt.3.html) here and maybe also [pam_setcred(pamh, PAM_REFRESH_CRED)](https://linux.die.net/man/3/pam_setcred) after that.

bvr-yr:
how about displaying this messages?

image_2024-03-20_15-05-01

that's if i set `pam_module = hyprlock`, default one shipped
mine only differs by `auth include system-auth` because of autologin

currently input is blocked (which is okay) but it is stuck at `checkWaiting` i guess?

150510-area- 26 639-541x161

if i do
```sh
faillock --user bvr --reset
```
from another tty it does nothing, only killing with `USR1`
probably even separate color could be dedicated to `faillock` later

src/core/Auth.cpp
PaideiaDilemma: Cool that you learned smth. I learned that pam is a bit of a mess ^^.

I don't think we want `pam_acct_mgmt`.
That is because the user that locks the screen has to unlock it. It is not the job of the lockscreen to check if that user's account is valid.

Similiar thing with `pam_set_cred`. Irrelevant for a lockscreen.

Greetd has to probably care about that but not us luckily :)

PaideiaDilemma:
Displaying if the account gets locked would be nice I think. But not in this PR probably.

I assume that you previously doing `faillock --user bvr --reset` worked fine right? I will try to repro and fix that.

Also the default `pam_module` should of be the one we ship. I just wanted to check if it works on most of the major distros, because this PR removes the fallback.

Thanks for the input and testing!

PaideiaDilemma:
I was not able to directly reproduce your problem @bvr-yr.

But I fixed some issues that might have caused it beeing stuck at checkWaiting.
On a system I tested it on I was able to input passwords normally when faillock locked the account. And I was also able to reset faillock from a different tty.

The Invalid Key down logs are currently a problem when switching to a different tty.
Then hyprlock looses focus and the key up event is not getting registered leading to those Logs. Unrelated to this PR.

Regarding displaying if the account is locked:
Looked into it a bit and its not that easy to check if your account has been locked via `faillock_pam`. But you can use a command like.
```sh
test "$(faillock --user $my_user | wc -l)" -ge "$max_attempts_plus_two" && echo "Locked"`
```

bvr-yr:
yeah can confirm now works correctly, unlocking by `sleep 60 && faillock --reset`, even after switching to another tty and reset as root I was able to type and unlock, although `Invalid key down` logged anyway
nice!

PaideiaDilemma:
closed by accident xD.
Tested some 2fa stuff today. Works fine after fixing some stuff I overlooked.

https://github.com/hyprwm/hyprlock/assets/78690852/15147fd3-edf0-441f-ae20-205d5fab3f8a

niksingh710:
> closed by accident xD. Tested some 2fa stuff today. Works fine after fixing some stuff I overlooked.
> hyprlock_2fa.mp4

how you configured this 2fA by using pam can you give some insight or direct me to the docs?

PaideiaDilemma:
This is [google-authenticator-libpam](https://github.com/google/google-authenticator-libpam).
The repo gives some instructions to set it up.

niksingh710:
@PaideiaDilemma Thanks for linking it.
This seems to be nice will give that a shot.

meanwhile i have fingerprint in my laptop.

And this is my pam file for `hyprlock`
```bash
auth      sufficient pam_fprintd.so
auth        include     login
``` 
currently I need to press enter on a blank input of the password, and then it takes my fingerprint and unlocks

will this pr resolve that or need to create a new issue entry?

PaideiaDilemma:
This PR should make it so that you don't have to press enter. You could test this by building from source. Would be good to know if it works properly with `pam_fprintd.so` (meaning that fingerprint works and you can still enter your password as usual)

Rattlyy:
@PaideiaDilemma actually, fingerprint works if you use `pam-fprint-grosshack`. My pam file, taken from the Arch Wiki:
```
auth		sufficient  	pam_fprintd_grosshack.so
auth		sufficient  	pam_unix.so try_first_pass nullok
```


https://github.com/hyprwm/hyprlock/assets/40183121/37cec3e8-b878-44fd-8710-5e22369fec6f

PaideiaDilemma:
Ahh ok that is nice. Did you give it a try with normal fprintd as well?

Curious if it has like a prompt or something.
My laptop has a goodix 538d and i did not get that to work, so i couldn't try it out.

Rattlyy:
> Ahh ok that is nice. Did you give it a try with normal fprintd as well?
> 
> Curious if it has like a prompt or something. My laptop has a goodix 538d and i did not get that to work, so i couldn't try it out.

That is normal fprint. It just adds the PAM library needed to run that. I personally have an elanmoc 0c00 and i'm using libfprintd-elanmoc2-newdrvs too.

PaideiaDilemma:
Ok yeah I meant with just fprintd. I would like to know if there is a prompt (This PR introduces "$PROMT", which you can add to you input:placeholder_text), if this PR addresses https://github.com/hyprwm/hyprlock/issues/250 and if the normal password input works as expected. Would appreciate if you could check some of that :)

Rattlyy:
$PROMPT works but sometimes it just becomes empty, in the vid I hid the
input as I don’t like it but with the fprintd-pam-grosshack I can input
both password and fingerprint.

Il giorno mar 2 apr 2024 alle 23:37 Maximilian Seidler <
***@***.***> ha scritto:

Rattlyy:
@PaideiaDilemma 
Any way to display this in the input field?
`[ERR] PAM: Failed to match fingerprint`

PaideiaDilemma:
Yes we could, but not in this PR.

First I have to figure out your report about
> $PROMPT works but sometimes it just becomes empty

and also if @vaxerski (pls excuse the ping ^^) would even want to merge this PR or not.

vaxerski:
I don't have much against this mr

bvr-yr:
let's go then and new release :rocket:

vaxerski:
tag me when this is ready for review

PaideiaDilemma:
Idk the Prompt being empty might be related to a weird pam config. Should not happen normally, since we are not setting a new placeholder when the prompt text is empty.

I think it is ready @vaxerski

Will make a wiki mr later today!

vaxerski:
alr wiki mr then

PaideiaDilemma:
@vaxerski done

PaideiaDilemma:
Nevermind I just reproduced the empty placeholder and noticed I did some dumb stuff. let me check that again. :sweat_smile:

vaxerski:
tag me when done then

PaideiaDilemma:
Alright @vaxerski ready for review

Reason I needed to revision some stuff is that I noticed that not restarting the authentication immediately on a failure was dump. Now `m_bDisplayFailText` handles how long the error message is displayed and that allows us to just block the input as long the auth thread is not waiting for input. That simplified some stuff that was ugly before.

I tested multiple authentication stages, as well as the weird su pam config on fedora.

src/core/hyprlock.cpp
```diff
@@ -379,6 +379,9 @@ void CHyprlock::run() {
 
     acquireSessionLock();
 
+    g_pAuth = std::make_unique<CAuth>();
+    g_pAuth->start();
```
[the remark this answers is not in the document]
PaideiaDilemma: No but it is actually also required if we want to display the pam prompt right away

5ysk3y:
Edge case, but would it also be possible to allow the setting of the path for the PAM module directory that is checked here or does that have to be hardcoded? I noticed from your wiki PR: `If the module isn't found in /etc/pam.d, "su" will be used as a fallback` - I run on NixOS and it wont store certain PAM modules in that particular directory, e.g. if I want to use u2f auth (yubikey), as its an extra package:

```
$ find / -name 'pam_u2f.so' 2>/dev/null
/nix/store/83rbp1sxjhpg9gdlzclnjsng5a0n7cq9-system-path/lib/security/pam_u2f.so
/nix/store/azg8kfwr70fw7x4fisidkilazai2naw0-system-path/lib/security/pam_u2f.so
/nix/store/gj8913xflvfms77df6y3cp5fv39q94hp-system-path/lib/security/pam_u2f.so
``` 

If I declare `pam_module` as `pam_u2f.so` I just get this from Hyprlock because `/etc/pam.d` doesnt have the module:

`[ERR] Pam module "pam_u2f.so" not found! Falling back to "su"`

If we can delcare the path then when the nix module is updated the path can also be set more in a more nix-friendly manner through that? Not sure if that would be in scopre for this PR.

PaideiaDilemma:
When you add a pam configuration via NixOs, it gets added to /etc/pam.d.
`pam_u2f.so` is a shared lib that can be used within a pam configuration. please look at [nixos option search for pam](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=security.pam.services.%3Cname%3E)

5ysk3y:
> When you add a pam configuration via NixOs, it gets added to /etc/pam.d. `pam_u2f.so` is a shared lib that can be used within a pam configuration. please look at [nixos option search for pam](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=security.pam.services.%3Cname%3E)

You're better than me ;) Let me take a look, nice one.

src/core/hyprlock.cpp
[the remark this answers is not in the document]
PaideiaDilemma: when we call start it will start the conversation and call `waitForInput`, which will block until we submit input.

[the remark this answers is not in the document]
vaxerski: oke my bad

vaxerski (approved):

vaxerski:
top

vaxerski:
@PaideiaDilemma can I get a wiki mr for the auth opt

PaideiaDilemma:
Here @vaxerski https://github.com/hyprwm/hyprland-wiki/pull/582

Mikilio:
Apparently there it is a known issue with pam_fprintd.so that you can't disable the timeout. There is a [PR](https://gitlab.freedesktop.org/libfprint/fprintd/-/merge_requests/195) for that.

For now, there should be a mechanism to test user presence before starting auth. Something like press any key to unlock. Something that can easily be added and removed via config.

bvr-yr:
[this](https://github.com/hyprwm/hyprlock/pull/115#issuecomment-2079225518) keeps troubling me, I was getting faillocks pretty often cause I use hyprlock with grace when smth is playing.
Personally I just added `faillock --reset` since it anyway launched through script, but still

I tried to dig a bit and it seems like `pam_end` never happens on grace end, as well as on `kill -USR1`, although it looks like it should.
But again, with conv returning `PAM_CONV_ERR` which will fail anyway I suppose? (Ok for `USR1` but not for grace unlock I think)
Can you clarify please, that's a culprit or something wrong with pam on me end?

PaideiaDilemma:
I do not understand why faillock would be involved at all, when a grace unlock happens.
When grace happens, then the main loop calls `g_pAuth->terminate();`

Can you open an issue for that?
Sadly I currently don't have a lot of time, but I will try to look into it if I could reproduce it.
I also want to bring to make faillock displayable in hyprlock, like suggested in this PR.

Issue #137, opened 2024-03-06
[REQ] Using a monitor's description for monitor value


matt1432:
If I try to do this:
```
input-field {
  monitor = desc:GIGA-BYTE TECHNOLOGY CO. LTD. G27QC 0x00000B1D
```

it won't actually put the input-field on that monitor.

This exact monitor value works on Hyprland

vaxerski:
ye thats because this is not hyprland and its not impl'd yet xd

matt1432:
fair enough lol, consider this a feature request then!

PR #215, opened 2024-03-21
config: allow hyprpaper as path to get backgrounds

bvr-yr:
this is a rough try, I don't know if it's really needed, just an idea from Discord

probably useful if hyprpaper used like dynamic changer

<s>I think it's mostly ready.</s>

EDIT: forgot about mon description, so probably after #137
EDIT2: actually seems to work now with mon description in hyprpaper.conf

Just the question: do we need this?

### DIFF
--- a/src/config/ConfigManager.hpp
+++ b/src/config/ConfigManager.hpp
@@ -21,10 +21,11 @@ class CConfigManager {
         std::unordered_map<std::string, std::any> values;
     };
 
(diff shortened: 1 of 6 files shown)


vaxerski:
> Just the question: do we need this?

uhhh... no?

bvr-yr:
> > Just the question: do we need this?
> 
> uhhh... no?

yeeaah.. kind of
alrighty then

PR #232, opened 2024-03-25
Nix/HM module: add sources and extraConfig

### LINKED ISSUE #231: [REQ] Add the ability to source config files in nix home manager module
I would like to source external files while using the home manager module, but I cannot find a way to do so. I do not see any `source` option, nor an `extraConfig` option I could fall back upon.

fufexan:
Fixes #231

@Anomalocaridid can you test whether this fixes your issue?


Anomalocaridid:
It works great! Thanks!

PR #234, opened 2024-03-26
widgets: add `rotate` option to label and image

bvr-yr:
self-explanatory

220517-output- eDP-1

proper way would be to do it directly with pango/cairo with any angle, and I probably will rework when I figure out how (particularly getting new rotated extents)

### DIFF
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -16,25 +16,31 @@ namespace std {
 }
 #endif
 
-Vector2D IWidget::posFromHVAlign(const Vector2D& viewport, const Vector2D& size, const Vector2D& offset, const std::string& halign, const std::string& valign) {
+Vector2D IWidget::posFromHVAlign(const Vector2D& viewport, const Vector2D& size, const Vector2D& offset, const std::string& halign, const std::string& valign, const double& ang) {
+
+    // offset after rotation for alignment
+    Vector2D rot;
+    if (ang != 0)
+        rot = (size - size.rotated(ang)) / 2.0;
+
     Vector2D pos = offset;
     if (halign == "center")
         pos.x += viewport.x / 2.0 - size.x / 2.0;
     else if (halign == "left")
-        pos.x += 0;
+        pos.x += 0 - rot.x;
     else if (halign == "right")
-        pos.x += viewport.x - size.x;
+        pos.x += viewport.x - size.x + rot.x;
     else if (halign != "none")
         Debug::log(ERR, "IWidget: invalid halign {}", halign);
 
     if (valign == "center")
         pos.y += viewport.y / 2.0 - size.y / 2.0;
     else if (valign == "top")
-        pos.y += viewport.y - size.y;
+        pos.y += viewport.y - size.y + rot.y;
     else if (valign == "bottom")
-        pos.y += size.y;
+        pos.y += 0 - rot.y;
     else if (valign != "none")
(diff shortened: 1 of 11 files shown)


vaxerski (commented):
are you dumb

vaxerski:
take a closer look at `wlr_matrix_project_box` private

bvr-yr:
what's wrong? I'm dumb

vaxerski:
look at the dam `wlr_matrix_project_box` function

bvr-yr:
you mean rotation there?
okay, but u said don't touch render function :D

vaxerski:
you can for this

vaxerski:
it just was related to the previous mr

bvr-yr:
yeah alright, will do it tomorrow
maybe do the same for Image too?

vaxerski:
fair

bvr-yr:
I'm struggling here

that works bad with rounded corners
only works when ratio 1:1 and full circle if rotated
otherwise I get this

image

image

so I added some checks

also made some changes to align labels and images properly to left/right top/bottom

considering all this, I have **strong feeling** that smth is missing

like.. matrix stuff not respected when performing rounding? or else

src/renderer/widgets/IWidget.cpp
```diff
@@ -16,25 +16,31 @@ namespace std {
...
     else if (halign != "none")
         Debug::log(ERR, "IWidget: invalid halign {}", halign);
 
     if (valign == "center")
         pos.y += viewport.y / 2.0 - size.y / 2.0;
     else if (valign == "top")
-        pos.y += viewport.y - size.y;
+        pos.y += viewport.y - size.y + rot.y;
     else if (valign == "bottom")
-        pos.y += size.y;
```
bvr-yr, line 35: take a look at discussion here #235 
as @PaideiaDilemma said

> Yeah so I think it does not make sense to have both `none` and `bottom`. But same with `none` and `left` I guess. But it behaves as you would expect with `left`.

this means it could be minor breakage for peeps who already use negative offset for bottom alignment, but I have to do this to properly align when rotated

vaxerski:
dear rizzler

when rotating, you have to render the object first onto a framebuffer the size of it, then render the texture of that fb with the rotation

regards,
vaxry

bvr-yr:
okay, totally not sure what I am doing but seems working :D

vaxerski:
nope since label isnt rounded

src/renderer/widgets/Label.cpp
```diff
@@ -111,12 +112,15 @@ bool CLabel::draw(const SRenderData& data) {
...
             shadow.markShadowDirty();
         }
     }
 
+    // calc pos
+    pos = posFromHVAlign(viewport, asset->texture.m_vSize, configPos, halign, valign, angle);
+
     CBox box = {pos.x, pos.y, asset->texture.m_vSize.x, asset->texture.m_vSize.y};
+    box.rot  = angle;
     shadow.draw(data);
```
vaxerski, line 124: dont you need to pass rot to shadow too?

bvr-yr: mmm not sure, seems all apply correctly like so, same as if I move it upper

image


do it?

vaxerski: if it works it works whatever

src/renderer/widgets/IWidget.cpp
vaxerski: meh

vaxerski (approved):
lgtm

PR #235, opened 2024-03-26
widget: fix valign bottom

PaideiaDilemma:
no need to correct for the size since bottom is 0

people probably already have some negative values for widgets that at the bottom. they would need to adjust.

### DIFF
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -32,7 +32,7 @@ Vector2D IWidget::posFromHVAlign(const Vector2D& viewport, const Vector2D& size,
     else if (valign == "top")
         pos.y += viewport.y - size.y;
     else if (valign == "bottom")
-        pos.y += size.y;
+        pos.y += 0;
     else if (valign != "none")
         Debug::log(ERR, "IWidget: invalid halign {}", halign);
 
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -34,7 +34,7 @@ Vector2D IWidget::posFromHVAlign(const Vector2D& viewport, const Vector2D& size,
     else if (valign == "bottom")
         pos.y += 0;
     else if (valign != "none")
-        Debug::log(ERR, "IWidget: invalid halign {}", halign);
+        Debug::log(ERR, "IWidget: invalid valign {}", valign);
 
     return pos;
 }


bvr-yr:
something tells me it is intended, just counter-intuitive
`none` is kinda true bottom

PaideiaDilemma:
are you sure? this is how it looks with `valign="bottom"` and `y=0` (the `asdf` label)

bvr-yr:
> are you sure? this is how it looks with valign="bottom" and y=0 (the asdf label)

yeah I know how it works now. I don't know, just a guess

> Also "none" is invalid valign no? Litterally prints an error.

hm, don't see any errors, it's even documented in Wiki as a valid value

src/renderer/widgets/IWidget.cpp
```diff
@@ -32,9 +32,9 @@ Vector2D IWidget::posFromHVAlign(const Vector2D& viewport, const Vector2D& size,
     else if (valign == "top")
         pos.y += viewport.y - size.y;
     else if (valign == "bottom")
-        pos.y += size.y;
+        pos.y += 0;
     else if (valign != "none")
-        Debug::log(ERR, "IWidget: invalid halign {}", halign);
+        Debug::log(ERR, "IWidget: invalid valign {}", valign);
```
PaideiaDilemma, line 37: I mean this error

PaideiaDilemma: but you are right, it is actually listed as valid in the wiki xD

PaideiaDilemma:
We will see what vaxry has to say about it ^^

src/renderer/widgets/IWidget.cpp
PaideiaDilemma: Uhm I am sorry it is actually `!=`. kinda missed that.
Yeah then you are right @bvr-yr and this is probably intended.

bvr-yr: but what's wrong here, it's `!= "none"` ?
shouldn't print if `none`

bvr-yr:
wait, don't rush xD
maybe you are right

src/renderer/widgets/IWidget.cpp
PaideiaDilemma: yeah if it is `none` it wont print.
I missed that `none` means "don't align" and overlooked that it is a valid value.

PaideiaDilemma:
Don't think I am right. Kinda rushed creating this PR as well so I guess that checks out.

bvr-yr:
but you are definitely right that it's confusing
didn't intend to choke you :D

PaideiaDilemma:
No worries, thanks for enlighting me.

Yeah so I think it does not make sense to have both `none` and `bottom`. But same with `none` and `left` I guess. But it behaves as you would expect with `left`. 

So maybe I am right after all :sweat_smile:

PR #247, opened 2024-03-30
image: add reload options

### LINKED ISSUE #225: [Req] Reload the image widget
Hi! It would be cool if we can reload the image on the widget. Or just by querying again the image on the path. 

Don't have any idea if this is feasible, but the use case for example is having a media player playing and showing the thumbnail of the $artURL of the music. 

image
This interests me more, but I think the SIG only takes effect for label widgets.

bvr-yr:
close #225

- `reload_time` -  interval in secs to reload, if 0 can be reloaded with `SIGUSR2`
- `reload_cmd` - command to obtain new path, if empty reuse default one

not sure with option names (as always)

considering that only png is supported, I guess main usecase should be converting and then `killall -USR2` in script, if album arts for example

https://github.com/hyprwm/hyprlock/assets/130279855/5e910e96-ca8a-428f-b27e-4d422a7cab6e

mostly copied from label widget

### DIFF
--- a/src/renderer/widgets/Image.cpp
+++ b/src/renderer/widgets/Image.cpp
@@ -1,6 +1,7 @@
 #include "Image.hpp"
 #include "../Renderer.hpp"
 #include "../../core/hyprlock.hpp"
+#include <filesystem>
 #include <cmath>
 
 CImage::~CImage() {
@@ -20,8 +21,15 @@ static void onAssetCallback(void* data) {
     PIMAGE->renderSuper();
 }
 
+std::uintmax_t CImage::getFileSize(const std::string& path) {
+    if (std::filesystem::exists(path))
+        return std::filesystem::file_size(path);
+
+    return fileSize;
+}
+
 std::string CImage::getUniqueResourceId() {
-    return std::string{"image:"} + std::to_string((uintptr_t)this) + ",path:" + path + ",time:" + std::to_string(std::chrono::system_clock::now().time_since_epoch().count());
+    return std::string{"image:"} + path + ",size:" + std::to_string(fileSize);
 }
 
 void CImage::onTimerUpdate() {
@@ -38,6 +46,11 @@ void CImage::onTimerUpdate() {
         if (path.empty())
             return;
     }
+
+    if (getFileSize(path) == fileSize)
+        return;
+
+    fileSize = getFileSize(path);
 
     // request new
     request.id        = getUniqueResourceId();
@@ -75,7 +88,8 @@ CImage::CImage(const Vector2D& viewport_, COutput* output_, const std::string& r
     reloadTime    = std::any_cast<Hyprlang::INT>(props.at("reload_time"));
     reloadCommand = std::any_cast<Hyprlang::STRING>(props.at("reload_cmd"));
 
-    angle = angle * M_PI / 180.0;
+    fileSize = getFileSize(path);
+    angle    = angle * M_PI / 180.0;
 
     plantTimer();
 }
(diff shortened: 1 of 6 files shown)


src/renderer/widgets/Image.cpp
```diff
@@ -1,7 +1,64 @@
...
+
+        if (path.ends_with('\n'))
+            path.pop_back();
+
+        if (path.empty())
+            return;
+    }
+
+    // request new
+    request.id        = getUniqueResourceId();
```
vaxerski, line 43: maybe fuckin stat the file first and check if it has actually changed instead of reloading it all the time??

bvr-yr: yeah a bit braindead rushed, I think just filesize is enough? what's the probability of same name + same size but different content

also don't think there is need to unload them at all, might be reused?

vaxerski: just stat the modification date bro

> also don't think there is need to unload them at all, might be reused?

unload old when loading new.

src/renderer/widgets/Image.cpp
```diff
@@ -54,6 +150,8 @@ bool CImage::draw(const SRenderData& data) {
         borderBox.round();
         imageFB.alloc(borderBox.w, borderBox.h, true);
         g_pRenderer->pushFb(imageFB.m_iFb);
+        glClearColor(0.0, 0.0, 0.0, 0.0);
+        glClear(GL_COLOR_BUFFER_BIT);
```
bvr-yr, line 154: can this help with #256? 
though I can't repro, pretty sure it has smth to do with framebuffer
maybe something more needed?

vaxerski: ah yea it should

vaxerski (approved):

bvr-yr:
nohow, it's processed cava raw output, you can check this python [example](https://github.com/karlstav/cava/issues/123#issuecomment-307891020)

particularly what you see on video, is futher processing of array (bar lengths) to print `|` character

mahaveergurjar:
> nohow, it's processed cava raw output, you can check this python [example](https://github.com/karlstav/cava/issues/123#issuecomment-307891020)
> 
> particularly what you see on video, is futher processing of array (bar lengths) to print `|` character

I only need cava when run  music and also need hide the all things when music run only see the art and cava 
240415_20h50m08s_screenshot
 Add this time my lock see this type

bvr-yr:
> I only need cava when run music and also need hide the all things when music run only see the art and cava

you need a wrapper script that will launch hyprlock with different config when something is playing, like
```sh
#!/usr/bin/bash

if [[ $(playerctl status) == Playing ]]; then
    hyprlock --config ~/.config/hypr/music.conf
else
    hyprlock
fi
```

mahaveergurjar:
Please Can you send music.conf with cava

bvr-yr:
just use Shape with same size and position you will launch cava at, nothing special
except Y position is bottom-based in hyplock config

for example, 1920x1080 monitor:

```sh
hyprctl dispatch -- exec "[float; size 600 200; move 660 740; noborder; noshadow; rounding 0] SDL_VIDEODRIVER=wayland SDL_VIDEO_ALLOW_SCREENSAVER=1 cava"
```
before launching hyprlock, and
```ini
shape {
    size = 600, 200
    position = 660, 140
    halign = none
    valign = none
    xray = true
}
```

I also suggest setting `method = sdl_glsl` in output section of cava's config

mahaveergurjar:
Thanks

PR #248, opened 2024-03-30
widgets: add user description variable

mrivnak:
Adds a `$DESC` variable that populates with the user GECOS field, which is commonly used for a display name in login managers, etc.

e.g. 
$USER = linux-user
$DESC = Linux User


vaxerski:
wiki mr pls

vaxerski (approved):

PR #257, opened 2024-04-05
core: avoid double unlocks with no_fade_out

PaideiaDilemma:
Fix for releasing the session lock twice when `general:no_fade_out` and when unlocking via signal


vaxerski (approved):
lgtm

PR #266, opened 2024-04-09
widgets: add Shape + minor image improve

### LINKED ISSUE #253: [REQ] Add custom shape
Some users (aka me, nobody cares) may want to fill a custom rectangle as background of labels, input, whatever, to better distinguish them from the all other stuff. At the moment this can be done with an image (like picture below), but this has disadvantages:
- You need a .png picture of solid color, want to change it - go to gimp, ugh
- Picture must be with specific ratio, same as above

2024-04-01T12:00:06,417336041+04:00

It would be easier, if there will be a widget, that just fill rectangle of custom shape with specified color, that also has border, shadow and all eye candy stuff. Also it must lay behind other widgets - images, labels etc. Basically something like this:

```
shape (or a rectangle, box, doesn't matter how it called) {
    monitor =
    color = rgba(14881337) # background color
    size = 228, 228 # x, y
    rounding = 69 # negative values mean circle
    border_size = 1
    border_color = rgb(696969)
    rotate = 0 # degrees, counter-clockwise

    # All shadow stuff ...

    position = 0, 0
    halign = center
    valign = center
}
```

bvr-yr:
that's to close #253

nothing special, except xray

well, despite it's stupid to punch a hole in background, someone may find it useful

I've been messaged to share a config from #247 video for cava bars, and I think updating labels every 33 ms is much more stupid :D

anyways, it is a harmless option after all

https://github.com/hyprwm/hyprlock/assets/130279855/53ef3f14-563d-4b75-bd28-23c0a85e246b

dunno, looks ok

also added some minor stuff to Image

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -179,6 +179,95 @@ in {
       ];
     };
 
+    shapes = mkOption {
+      description = "Shape configurations";
+      type = listOf (submodule {
+        options = {
+          monitor = mkOption {
+            description = "The monitor to draw a shape";
+            type = str;
+            default = "";
+          };
+
+          size = {
+            x = mkOption {
+              description = "Width of the shape";
+              type = int;
+              default = 360;
+            };
+            y = mkOption {
+              description = "Height of the shape";
+              type = int;
+              default = 60;
+            };
+          };
+
+          color = mkOption {
+            description = "Color of the shape";
+            type = str;
+            default = "rgba(22, 17, 17, 1.0)";
+          };
+
+          rounding = mkOption {
+            description = "Rounding of the shape";
+            type = int;
+            default = -1;
+          };
+
+          border_size = mkOption {
+            description = "Size of shape border";
+            type = int;
+            default = 4;
+          };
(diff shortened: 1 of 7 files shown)


Ligthiago:
Nice. 
But I encounter a bug: if the  height of shape smaller or equal to width (size = 400, 300 for example) and shape is rotated - I get weird artifact or borders, see picture. Optionally you have to set a big `rounding` to see a bug better

Similar to this has been on images before get fixed.

2024-04-09T16:32:12,661813850+04:00

bvr-yr:
aye just wanted to ask vaxry if I need to glClear here too, will add

vaxerski:
я тоже понял

vaxerski (approved):
lgtm
wiki mr pls

also where is the backdoor

bvr-yr:
maybe create FB no matter rotated or not? honestly looks a bit ugly on fade-in/fade-out when borderBox is bright and shape is dark

vaxerski (approved):
yah

Issue #263, opened 2024-04-08
hold backspace to delete text entered


kojq:
option to hold backspace to delete text entered instead of pressing the backspace key each time.

also, seems like the `delete` key is considered a character when it does not need to be.

nhosama:
> option to hold backspace to delete text entered instead of pressing the backspace key each time.

You can do `Esc`, `Ctrl + u` or `Ctrl + Backspace` though.

[https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock/#keyboard-shortcuts-and-actions](https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock/#keyboard-shortcuts-and-actions)

kojq:
That feature is not the same...

nhosama:
Oh! I misunderstood then.

PR #271, opened 2024-04-10
 hold backspace to delete text entered #263

kojq:
scuffed. need to fix.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -776,6 +776,7 @@ void CHyprlock::renderOutput(const std::string& stringPort) {
     PMONITOR->sessionLockSurface->render();
 }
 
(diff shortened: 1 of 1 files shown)


PaideiaDilemma:
I also drafted this one up.
You may look at it here: https://github.com/PaideiaDilemma/hyprlock/tree/hold-backspace

kojq:
I dunno how the generalized one would be used aside from when both entering and deleting text, but I like it.

Check if the repeatable rate is not too fast or slow (or could be configured). Other than that, I will close this when yours merges.

PaideiaDilemma:
> I dunno how the generalized one would be used aside from when both entering and deleting text, but I like it.

Honestly idk either, but it made more sense to me doing it like that.

kojq:
Closing as #277 is in good condition.

PR #277, opened 2024-04-11
core: add key repeat for backspace and make del clear the input

### LINKED ISSUE #263: hold backspace to delete text entered

PaideiaDilemma:
Allows users to hold backspace to delete the password buffer.
I feel like using a shortcut to clear the entire buffer is superior, but oh well.

A minor thing to consider:
If you press down a key, press&hold backspace after that and then release the first key, that will cancel the timer. Idk if we want to make sure that the key we are releasing is a repeatable one or not and only cancel if it is. That would fix that.

Also we don't necessarily need to move sym into a separate function, but it made more sense to me like that.

Closes #263

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -793,45 +793,23 @@ void CHyprlock::renderOutput(const std::string& stringPort) {
     PMONITOR->sessionLockSurface->render();
 }
 
-void onHoldBackspaceDeleteTimerCallback(std::shared_ptr<CTimer> self, void* data) {
-    g_pHyprlock->doBackspace();
-
-    for (auto& o : g_pHyprlock->m_vOutputs) {
-        o->sessionLockSurface->render();
-    }
-}
-
-void CHyprlock::doBackspace() {
-    if (m_sPasswordState.passBuffer.length() > 0) {
-        // handle utf-8
-        while ((m_sPasswordState.passBuffer.back() & 0xc0) == 0x80)
-            m_sPasswordState.passBuffer.pop_back();
-        m_sPasswordState.passBuffer = m_sPasswordState.passBuffer.substr(0, m_sPasswordState.passBuffer.length() - 1);
-    }
-}
-
 void CHyprlock::startKeyRepeat(xkb_keysym_t sym) {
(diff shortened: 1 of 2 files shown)


bvr-yr:
maybe add `Delete` to clear shortcuts instead of repeat? seems more handy and sensible

PaideiaDilemma:
Yeah you are right. Will do that later :)

JohnRTitor:
@vaxerski could you check this?

src/core/hyprlock.cpp
```diff
@@ -793,6 +793,29 @@ void CHyprlock::renderOutput(const std::string& stringPort) {
...
+        /* initial delay until key repeat */ std::chrono::milliseconds(500), [sym](std::shared_ptr<CTimer> self, void* data) { g_pHyprlock->repeatKey(sym); }, nullptr);
+}
+
+void CHyprlock::repeatKey(xkb_keysym_t sym) {
+    handleKeySym(sym);
+
+    // This condition is for backspace and delete keys, but should also be ok for other keysyms since our buffer won't be empty anyways
+    if (bool CONTINUE = m_sPasswordState.passBuffer.length() > 0; CONTINUE)
+        m_pKeyRepeatTimer = addTimer(
+            /* key repeat delay */ std::chrono::milliseconds(50), [sym](std::shared_ptr<CTimer> self, void* data) { g_pHyprlock->repeatKey(sym); }, nullptr);
```
vaxerski, line 812: you should get the values from wl_keyboard
image

src/core/hyprlock.hpp
```diff
@@ -82,6 +85,9 @@ class CHyprlock {
     xkb_keymap*                     m_pXKBKeymap  = nullptr;
     xkb_state*                      m_pXKBState   = nullptr;
 
+    int32_t                         m_iKeebRepeatRate  = 0;
```
vaxerski, line 88: sane defaults please, not all comps may send this info

vaxerski (commented):
rest ok

vaxerski (approved):
lgtm

PR #278, opened 2024-04-12
label: add `text_align` option for multi-line text

bvr-yr:
instead of messing with multiple labels with its' own alignments and shadow FBs, let pango do the job

there are more fancy stuff that could be added, like justify and ellipsize with explicit width and height, but I don't think it is relevant in context of a locker

or is it?

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -561,6 +561,12 @@ in {
               default = "Hi there, $USER";
             };
 
+            text_align = mkOption {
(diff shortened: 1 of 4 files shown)


bvr-yr:
I think there is not

not in here https://docs.gtk.org/Pango/pango_markup.html

and not a single `align` word in [src](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pango-markup.c)

I guess cause it is layout attr

vaxerski (approved):

PR #281, opened 2024-04-13
label: add `font_features` support

hlmos:
add a config option to apply stylistic sets to the `font_family`

config has a css sytax. eg `font_features = ss01, ss03` or `font_features = ss01 1, ss02 0, ss03 1`

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -579,6 +579,12 @@ in {
               default = "Noto Sans";
             };
 
+            font_features = mkOption {
+              description = "Font features to apply on the label";
+              type = str;
+              default = "";
+            };
+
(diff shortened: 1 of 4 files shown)


bvr-yr:
you can do it just inside `<span >`, no?

hlmos:
yup, but afaik there is no way to do it when the label is a command
this does break existing stylistic set config inside `<span >` though, so i would have to look into that

tho maybe the thing would be to allow combining markup and commands?

bvr-yr:
wdym? just use
```sh
echo -n "<span font_features='ss01'>$(command)</span>"
```
as cmd

hlmos:
true, great point

Issue #284, opened 2024-04-18
Possible trailing newline character from cmd output on label text


joelsleeba:
Hi, love the project.

I was trying to center the label text. It appears that same text given directly using `text = Sampletext` and `text = cmd[update:1000] echo Sampletext` changes the position in both instances.

Here's the lockscreen when using `text=Joel`
![image](https://github.com/hyprwm/hyprlock/assets/54855387/b6d90e5a-1b6b-44aa-8767-8b150f375f75)

And here's the lockscreen when using `text=cmd[update:1000] echo $USER`
![image](https://github.com/hyprwm/hyprlock/assets/54855387/5887e989-46b2-4e82-94f4-5eadf82a403c)

Based on the spacing added in the second image I guess the trailing newline character from the shell output is the culprit.

vaxerski:
likely

bvr-yr:
exactly. nothing wrong with that, `echo` without `-n` flag outputs newline char

joelsleeba:
Yes this works for `echo`. But for other commands I am currently piping through `head -c -1` to remove the trailing newline. 

I was under the impression that the trailing newline was not intentional. I will close the issue if otherwise

davc0n:
@vaxerski Hi, what do you think about an option to trim commands output? Eventually I can make a PR

bvr-yr:
I think you can always wrap cmd in
```bash
echo -n "$(cmd)"
```

vaxerski:
I don't care, you can make it into a mr

davc0n:
@joelsleeba @vaxerski I think you can close this.

joelsleeba:
Thank you for the fix

PR #310, opened 2024-05-04
renderer: add text trim with option to disable

davc0n:
ref. https://github.com/hyprwm/hyprlock/issues/284

Not sure if that's the best way to trim a string in C++, please let me know if there's a better solution.

If it will be accepted I'll update the wiki too.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -44,6 +44,7 @@ void CConfigManager::init() {
     m_config.addSpecialConfigValue(name, "shadow_boost", Hyprlang::FLOAT{1.2});
 
     m_config.addConfigValue("general:disable_loading_bar", Hyprlang::INT{0});
+    m_config.addConfigValue("general:disable_text_trim", Hyprlang::INT{0});
     m_config.addConfigValue("general:hide_cursor", Hyprlang::INT{0});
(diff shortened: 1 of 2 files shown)


src/config/ConfigManager.cpp
```diff
@@ -44,6 +44,7 @@ void CConfigManager::init() {
     m_config.addSpecialConfigValue(name, "shadow_boost", Hyprlang::FLOAT{1.2});
 
     m_config.addConfigValue("general:disable_loading_bar", Hyprlang::INT{0});
+    m_config.addConfigValue("general:disable_text_trim", Hyprlang::INT{0});
```
vaxerski, line 47: can you flip the option? (text_trim and default 1)

vaxerski (commented):
also needs a clang-format

src/config/ConfigManager.cpp
davc0n: the idea was to be coherent with other general options that are 0 by default, are you sure I should flip it?

vaxerski: yea, people complain about having both `xxx = bool` and `disable_yyy = bool` options and call it confusing. People generally prefer `enable_xxx` options it seems.

davc0n: done

vaxerski (approved):
lgtm thanks!

PR #326, opened 2024-05-13
Fixed getconf command in build instructions

pebble-24:
"getconf NPROCESSORS_CONF" isn't a valid command. The correct command is "getconf _NPROCESSORS_CONF"


vaxerski (approved):

fufexan:
We should use the Ninja generator and drop all the nproc stuff (as is done in the Hyprland Makefile).

PR #338, opened 2024-05-19
cmake: macro-prefix-map for __FILE__, generate compile_commands.json

rtgiskard:


vaxerski (commented):
what and why?

vaxerski (approved):
ok then, lgtm! Please next time write stuff in the MR description, I don't read commit comments.

PR #376, opened 2024-06-18
 fix second instance of hyprlock unlocking session

### LINKED ISSUE #373: Multiple instances of `hyprlock` should not be allowed to start
Thanks for creating `hyprlock` :)

I'm running it with `sway` and `swayidle` as follows:
```
exec swayidle \
    timeout 600 'hyprlock --quiet' \
    timeout 15 'if pgrep --exact hyprlock; then swaymsg "output * power off"; fi' resume 'swaymsg "output * power on"' \
    lock 'hyprlock --quiet --immediate' \
    before-sleep 'hyprlock --quiet --immediate' \
    after-resume 'swaymsg "output * power on"'
```
The idea behind this config:
* After 5 min of inactivity, lock the screen
* After 15 sec of inactivity, check if lock screen is running - if so, turn off displays (save my OLEDs from burning...)
* If commanded to do so by `loginctl` or system suspend, lock the screen immediately

Now the fun part - `hyprlock` can be started multiple times (and does nothing when started after the first instance btw) and that indeed happens when I command the lock. After the 5 min of inactivity, another instance of `hyprlock` is created. Now, when I unlock the screen I still have some `hyprlock` instance running in the background which I don't see - but I could spot it because after 15 sec my screens are turned off :smile:

`hyprlock` should have a "_locking mechanism_" (pun intended) to allow running only one instance.

Related: #351

thecodinglab:
Fixes #373

I noticed that the `onLockFinished` function calls `ext_session_lock_v1_unlock_and_destroy`, whereas Swaylock on the other hand directly calls `exit(2)` once it notices that a second lock is already running (see [this line](https://github.com/swaywm/swaylock/blob/b70d85ec145520ba3ebaa4c51921a2761bc0c7f0/main.c#L282)). Implementing this fix alone, however, did not resolve the problem. Thus, I dug further and noticed, that moving the call to `wl_display_roundtrip` after `acquireSessionLock` now seems to resolve the issue.

I have to admit that I have no idea about Wayland and their protocols, so please throw this code out of the window if it makes no sense at all.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -365,9 +365,6 @@ void CHyprlock::run() {
         exit(1);
     }
 
-    // gather info about monitors
-    wl_display_roundtrip(m_sWaylandState.display);
-
     g_pRenderer = std::make_unique<CRenderer>();
 
     const auto         CURRENTDESKTOP = getenv("XDG_CURRENT_DESKTOP");
@@ -392,6 +389,9 @@ void CHyprlock::run() {
     }
 
     acquireSessionLock();
+
+    // gather info about monitors
+    wl_display_roundtrip(m_sWaylandState.display);
 
     g_pAuth = std::make_unique<CAuth>();
     g_pAuth->start();
@@ -958,8 +958,7 @@ void CHyprlock::onLockLocked() {
 
 void CHyprlock::onLockFinished() {
     Debug::log(LOG, "onLockFinished called. Seems we got yeeten. Is another lockscreen running?");
-    ext_session_lock_v1_unlock_and_destroy(m_sLockState.lock);
-    m_sLockState.lock = nullptr;
+    exit(1);
 }
 
 ext_session_lock_manager_v1* CHyprlock::getSessionLockMgr() {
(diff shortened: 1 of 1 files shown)


src/core/hyprlock.cpp
```diff
@@ -393,6 +390,9 @@ void CHyprlock::run() {
 
     acquireSessionLock();
 
+    // gather info about monitors
+    wl_display_roundtrip(m_sWaylandState.display);
```
vaxerski, line 394: I dont think this is correct

thecodinglab: Yeah, I just noticed, that the second instance of hyprlock now segfaults. Initially, I moved the `acqureSessionLock` call onto line 368 (i.e. after the condition, but before `wl_display_roundtrip`), which also resolves the problem, though this introduces a brief red flashing screen, which does not happen when `wl_display_roundtrip` is called after the creation of the renderer.

vaxerski (approved):
sure I guess. Technically I believe we should destroy the lock object but it doesn't matter as we exit anyways

PR #380, opened 2024-06-24
Inproved error handling and arg parsing

JassonCordones:
Same as title

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -40,21 +40,25 @@ int main(int argc, char** argv, char** envp) {
         else if (arg == "--quiet" || arg == "-q")
             Debug::quiet = true;
 
-        else if ((arg == "--config" || arg == "-c") && i + 1 < (std::size_t)argc)
+        else if ((arg == "--config" || arg == "-c") && i + 1 < (std::size_t)argc) {
             if (auto value = parseArg(args, arg, i); value)
                 configPath = *value;
             else
                 return 1;
-        else if (arg == "--display" && i + 1 < (std::size_t)argc) {
+
+        } else if (arg == "--display" && i + 1 < (std::size_t)argc) {
             if (auto value = parseArg(args, arg, i); value)
                 wlDisplay = *value;
             else
                 return 1;
-        } else if (arg == "--immediate") {
+
+        } else if (arg == "--immediate")
             immediate = true;
-        } else if (arg == "--help" || arg == "-h") {
+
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -12,47 +13,83 @@ void help() {
...
+                return 1;
+            }
 
-        else if (arg == "--display" && i + 1 < argc) {
-            wlDisplay = argv[i + 1];
-            i++;
+        else if (arg == "--display" && i + 1 < static_cast<std::size_t>(argc)) {
+             if (auto value = parseArg(args, arg, i); value) {
+                wlDisplay = *value;
+            } else {
```
vaxerski, line 53: no {} around short ifs around here

src/main.cpp
```diff
@@ -12,47 +13,83 @@ void help() {
...
+            if (auto value = parseArg(args, arg, i); value) {
+                configPath = *value;
+            } else {
+                return 1;
+            }
 
-        else if (arg == "--display" && i + 1 < argc) {
-            wlDisplay = argv[i + 1];
-            i++;
+        else if (arg == "--display" && i + 1 < static_cast<std::size_t>(argc)) {
```
vaxerski, line 50: why not just use C-style casts? They are shorter.

src/main.cpp
```diff
@@ -12,47 +13,83 @@ void help() {
...
-    } catch (const char* err) {
-        Debug::log(CRIT, "ConfigManager threw: {}", err);
-        std::string strerr = err;
-        if (strerr.contains("File does not exist"))
+        auto configManager = std::make_unique<CConfigManager>(configPath);
+        configManager->init();
+        g_pConfigManager = std::move(configManager);
+    } catch (const std::exception& ex) {
+        Debug::log(CRIT, "ConfigManager threw: {}", ex.what());
+        if (std::string(ex.what()).find("File does not exist") != std::string::npos) {
```
vaxerski, line 79: .contains

src/main.cpp
JassonCordones: I just default to static_cast since they can catch wrong conversions during compile time, but I don't see need for that in this case. I can change it to C-style if you want.

vaxerski: yeah we usually use c style

JassonCordones:
Done.

vaxerski (commented):
rest ok

vaxerski (approved):
lgtm

Issue #357, opened 2024-06-02
Time, Date and other stuff aren't updating


ilolm:
I use Arch Linux with the Hyprland desktop environment. Hyprlock is working fine except for one issue: when I lock my machine and the PC is idle (no mouse or keyboard activity), the time, date, and other information displayed by Hyprlock do not update. They only update when I move my mouse or type something on my keyboard.

Please watch the attached video for a demonstration.

https://github.com/hyprwm/hyprlock/assets/83036869/62906154-9366-4349-a2b3-c832d2d7296f

ilolm:
SOLUTION:

Just installed aur/lib32-nvidia-utils-beta aur/nvidia-beta-dkms aur/nvidia-settings-beta aur/nvidia-utils-beta                        555.42.02-1

ashish-kus:
text = cmd[update:1000] echo -e "$(date +"%M")"

did you used this format.

ilolm:
```
# /* ---- 💫 https://github.com/JaKooLit 💫 ---- */  #
# Hyprlock                                                          
# Original config submitted by https://github.com/SherLock707

# Sourcing colors generated by wallust
source = $HOME/.config/hypr/wallust/wallust-hyprland.conf
$Scripts = $HOME/.config/hypr/scripts

general {
    grace = 1
}

background {
    monitor =
	#path = screenshot   # screenshot of your desktop
	path = $HOME/.config/rofi/.current_wallpaper   # NOTE only png supported for now
    color = $color7

    # all these options are taken from hyprland, see https://wiki.hyprland.org/Configuring/Variables/#blur for explanations
    blur_size = 6
    blur_passes = 2 # 0 disables blurring
    noise = 0.0117
    contrast = 1.3000 # Vibrant!!!
    brightness = 0.8000
    vibrancy = 0.2100
    vibrancy_darkness = 0.0
}

input-field {
    monitor = DP-1
    size = 250, 50
    outline_thickness = 3
    dots_size = 0.33 # Scale of input-field height, 0.2 - 0.8
    dots_spacing = 0.15 # Scale of dots' absolute size, 0.0 - 1.0
    dots_center = true
    outer_color = $color5
    inner_color = $color0
    font_color = $color12
    fade_on_empty = true
    placeholder_text = <i>Password...</i> # Text rendered in the input box when it's empty.
    hide_input = false
    
    position = 0, 200
    halign = center
    valign = bottom
}

# Date
label {
    monitor = DP-1
[...]

PR #384, opened 2024-06-26
core: stabilize label updates and revision locking in the asyncResourceGatherer

### LINKED ISSUE #367: Datarace in CAsyncResourceGatherer
Current implementation of `CAsyncResourceGatherer` is racy.

Data race between `preloadTargets` [modification](https://github.com/hyprwm/hyprlock/blob/main/src/renderer/AsyncResourceGatherer.cpp#L406) in `CAsyncResourceGatherer::asyncAssetSpinLock` and [accessing](https://github.com/hyprwm/hyprlock/blob/main/src/renderer/AsyncResourceGatherer.cpp#L215) in `CAsyncResourceGatherer::apply` leads to some widgets not shown. (It this case due to vector reallocation after push_back new element).

<details>

<summary>Hotfix:</summary>

```diff
diff --git a/src/renderer/AsyncResourceGatherer.cpp b/src/renderer/AsyncResourceGatherer.cpp
index f7f46ee..5f21650 100644
--- a/src/renderer/AsyncResourceGatherer.cpp
+++ b/src/renderer/AsyncResourceGatherer.cpp
@@ -211,6 +211,7 @@ void CAsyncResourceGatherer::gather() {
 }

 void CAsyncResourceGatherer::apply() {
+    std::lock_guard preloadTargetsLock{asyncLoopState.preloadTargetsMutex};

     for (auto& t : preloadTargets) {
         if (t.type == TARGET_IMAGE) {
@@ -400,6 +401,7 @@ void CAsyncResourceGatherer::asyncAssetSpinLock() {

         // process requests

+        std::lock_guard preloadTargetsLock{asyncLoopState.preloadTargetsMutex};
         asyncLoopState.busy = true;
         for (auto& r : requests) {
             if (r.type == TARGET_TEXT) {
diff --git a/src/renderer/AsyncResourceGatherer.hpp b/src/renderer/AsyncResourceGatherer.hpp
index 2fa81a8..01ffa1b 100644
--- a/src/renderer/AsyncResourceGatherer.hpp
+++ b/src/renderer/AsyncResourceGatherer.hpp
@@ -66,6 +66,8 @@ class CAsyncResourceGatherer {

         std::mutex                   assetsMutex;

+        std::mutex                   preloadTargetsMutex;
+
         std::vector<SPreloadRequest> requests;
         bool                         pending = false;
```

</details>

In other cases, for example, [accessing](https://github.com/hyprwm/hyprlock/blob/main/src/renderer/AsyncResourceGatherer.cpp#L86) non-atomic variable `asyncLoopState.busy` without additional synchronization leads to undefined behavior.

PaideiaDilemma:
Was able to repro #357 (multiple monitors required i think).
Fixes it on my end.

[4509552](https://github.com/hyprwm/hyprlock/pull/384/commits/450955217f860554384549f461cf0c080d86a957) reverts the previous approach of adding a timer to re-render, when the resource was not available (#152). It also sets `needsFrame=true`, when render is called but `frameCallback` is present (I think not 100% needed but still makes sense)

The other 3 commits make sure that `getAssetById` fails less often by removing the `busy` flag and revisioning the locking in the resource gatherer.

`-fsanitize=thread` is a bit happier regarding the asyncResourceGatherer now. Besides some races with `dmas` on startup I think.

Closes #367

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -129,8 +129,10 @@ static const wl_callback_listener callbackListener = {
 void CSessionLockSurface::render() {
     Debug::log(TRACE, "render lock");
 
-    if (frameCallback || !readyForFrame)
+    if (frameCallback || !readyForFrame) {
+        needsFrame = true;
         return;
+    }
 
     const auto FEEDBACK = g_pRenderer->renderLock(*this);
     frameCallback       = wl_surface_frame(surface);
--- a/src/renderer/AsyncResourceGatherer.cpp
+++ b/src/renderer/AsyncResourceGatherer.cpp
@@ -11,8 +11,6 @@
 #include "../helpers/Jpeg.hpp"
 #include "../helpers/Webp.hpp"
 
-std::mutex cvmtx;
-
 CAsyncResourceGatherer::CAsyncResourceGatherer() {
     asyncLoopThread = std::thread([this]() {
         this->gather(); /* inital gather */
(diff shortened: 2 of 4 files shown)


vaxerski (approved):
thanks!

PR #386, opened 2024-06-27
core: improve dynamic output handling

### LINKED ISSUE #117: Lock surfaces not being created for new monitors (red screen after unplugging and replugging a display)
When turning my monitor off and on while hyprlock is active, I'm present with a red screen, this is happening because hyprland is detecting the monitor as a new one when disconnected and connected again and hyprlock is not creating a lock surface for the monitor after being added -

In the log provided in the screenshot, the part before I turn my screen on and off is everything up to the third "framebuffer created" messages

As you can see, initially, hyprlock configures a surface for the monitor (in the line before the framebuffer created message)

After I turn off and turn on the display (which is detected as unplugging and plugging it) - after the framebuffer created messages, you can see that it detects the display, but doesn't create a lock surface on it, causing a red screen

### LINKED ISSUE #337: Invisible lock when unlocking from a disabled screen.
Sorry for the edit, i miss typed and for some reason github automatically posted this. In any case:

I disable my laptop display on lid close as I use it with an external monitor sometimes.
```
bindl=,switch:on:Lid Switch,exec,hyprctl keyword monitor "eDP-1, disable"
bindl=,switch:off:Lid Switch,exec,hyprctl keyword monitor "eDP-1, 1920x1080, 0x0, 1.2"
```
however with hyprlock when I open my lid (re-enabling the display), I don't see any hyprlock prompt and I need to type my password on some invisible lock screen before I can interact with my computer.

### reproduction steps
1. make sure hyprlock is installed and works
2. ideally use a laptop or somehow disable and re-enable your only display
3. open your display to see a new empty workspace, you cannot interact with anything unless you type your password in some invisible prompt like you would normally.

### Additional info
tested on Hyprland-git with Hyprlock-git.

PaideiaDilemma:
I reproduced some crashes with the lid switching the internal monitor on and off.
That is because the session lock surface is getting created via `handleDone` for a new output.
Even if a full display roundtrip is done, some events might arrive earlier than the output `done` event causing a crash because the `sessionLockSurface` pointer is null. Checking for the `sessionLockSurface` pointer fixes that. Other option would be to only add dynamic outputs to the `m_vOutputs` vector when the sessionLockSurface already exists

Then I noticed that the widgets for a surface are not removed, when it is destroyed (I think that caused #337). Sometimes the new `sessionLockSurface` happens to be the same pointer as the original one. And because the widgets map is keyed by those pointers it will reuse the widgets leading to some weird stuff.
```
[LOG]   | removed iface 60
[LOG] output 905814070 done
[LOG]   | got iface: wl_output v4
[LOG]    > Bound to wl_output v4
[LOG] output 61 done
[LOG] Creating a surface dynamically for output as we are already locked
[LOG] sessionLockSurface 0x36054330
[LOG] output 61 make null model null
[LOG] output 61 name HEADLESS-1
[LOG] output 61 description Headless output 1
[LOG] output 61 done
...
[LOG]   | got iface: wl_output v4
[LOG]    > Bound to wl_output v4
[LOG]   | removed iface 61
[LOG] output 62 make Najing CEC Panda FPD Technology CO. ltd model 0x002B
[LOG] output 62 name eDP-1
[LOG] output 62 description Najing CEC Panda FPD Technology CO. ltd 0x002B (eDP-1)
[LOG] output 62 done
[LOG] Creating a surface dynamically for output as we are already locked
[LOG] sessionLockSurface 0x36054330
```

Stuff that is still to be investigated:
- DMA Frames are identified by output stringPorts, so when someone would unplug one monitor and plug another one in with different dimensions, that would cause weirdness probably. Should we identify them by description or something instead so that we don't need to remove them, or just remove existing ones when an output is removed?
- (Hyprland) My active workspace switches when I deactivate and reactivate my only output. I can imagine that is kind of supposed to happen, but I will need to check what is actually going on.
- (Hyprland) There is a red screen flashing, when reactivating a monitor. Also happens with swaylock. But not when replugging and external monitor.

Closes #337
Probably Closes #117

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -27,6 +27,8 @@ static const wp_fractional_scale_v1_listener fsListener = {
 };
 
 CSessionLockSurface::~CSessionLockSurface() {
+    g_pRenderer->removeWidgetsFor(this);
+
     if (fractional) {
         wp_viewport_destroy(viewport);
         wp_fractional_scale_v1_destroy(fractional);
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -27,8 +27,6 @@ static const wp_fractional_scale_v1_listener fsListener = {
 };
 
 CSessionLockSurface::~CSessionLockSurface() {
-    g_pRenderer->removeWidgetsFor(this);
-
     if (fractional) {
         wp_viewport_destroy(viewport);
         wp_fractional_scale_v1_destroy(fractional);
--- a/src/core/Output.cpp
+++ b/src/core/Output.cpp
@@ -25,11 +25,9 @@ static void handleDone(void* data, wl_output* output) {
     const auto POUTPUT = (COutput*)data;
     Debug::log(LOG, "output {} done", POUTPUT->name);
     if (g_pHyprlock->m_bLocked && !POUTPUT->sessionLockSurface) {
-        // if we are already locked, create a surface dynamically after a small timeout
-        // we also need to request a dma frame for screenshots
+        // if we are already locked, create a surface dynamically
         Debug::log(LOG, "Creating a surface dynamically for output as we are already locked");
         POUTPUT->sessionLockSurface = std::make_unique<CSessionLockSurface>(POUTPUT);
-        g_pRenderer->asyncResourceGatherer->recheckDMAFramesFor(POUTPUT);
     }
 }
 
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -570,9 +570,7 @@ void CHyprlock::unlock() {
     m_tFadeEnds    = std::chrono::system_clock::now() + std::chrono::milliseconds(500);
     m_bFadeStarted = true;
 
-    for (auto& o : m_vOutputs) {
-        o->sessionLockSurface->render();
-    }
+    renderAllOutputs();
 }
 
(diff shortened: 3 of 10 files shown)


vaxerski:
> or just remove existing ones when an output is removed?

prolly.

> My active workspace switches when I deactivate and reactivate my only output. I can imagine that is kind of supposed to happen, but I will need to check what is actually going on.

hl quirk.

> There is a red screen flashing, when reactivating a monitor. Also happens with swaylock. But not when replugging and external monitor.

hl bug.

PaideiaDilemma:
Idk how dynamic dma frames can work. 
The workspace below would need to flash for the screencopy to copy what you would want to blur.
Currently when a dynamic dma frame is created hyprlock itself gets screencopied making the new background a blur of existing widgets.

I think it might be better to just not support dynamic dma frames at all. But if one for a stringPort already exists, check if the dimensions match the ones of the new output. Otherwise fall back to the background color.

But maybe you have an idea how to make it work properly?

vaxerski:
right, you're right. Would need an extension or something. Fair. Let's not.

PaideiaDilemma:
Alright done.

What we don't support yet is dynamically updating widgets, when mode/scale of a monitor changes for some reason. I guess we could destroy the widgets in that case and let them be recreated. But thats for another mr.

vaxerski (approved):
thanks!

PR #393, opened 2024-07-02
input-field: add font_family

julianmarmier:
added font family customization for the input field, or was it intentional to leave it uncustomizable?


vaxerski:
you can use pango markup

PR #394, opened 2024-07-03
asyncResourceGatherer: add dma frames before starting the asyncLoopThread

PaideiaDilemma:
Fixes a race that now shows itself since https://github.com/hyprwm/hyprlock/pull/386, because if the added check to fall back to background color for dynamic backgrounds.

Thats because `gather` waits for all dma frames to be ready, but rarely it was possible that not all dma frames have been added yet in the `asyncResourceGatherer` ctor.


vaxerski (approved):
thanks!

PR #395, opened 2024-07-03
Remove dupe input-field rounding from listKeysForSpecialCategory

davc0n:
Very small PR, removes a code line which seems redundant, likely a copy paste typo.


vaxerski (approved):
yup, thanks

PR #396, opened 2024-07-04
label: do not error when using cmd[] for a static label cmd

PaideiaDilemma:
Currently `cmd[]` will cause the following error `[ERR] Unknown prop in string format`.

Since I think it makes sense to use it like that for a static label command, I set `removeEmpty` for the `CVarList` ctor.
Then I noticed `removeEmpty` does not affect empty inputs, so I figured It would make sense if it did. Thus https://github.com/hyprwm/hyprutils/pull/7.


vaxerski (approved):
thanks!

PR #398, opened 2024-07-05
config: don't crash when getenv HOME returns null

### LINKED ISSUE #392: Hyprlock is core dumping when running as a per-user system service
I know of hypridle which can handle locking and suspending the computer after a timeout but I wanted to use the systemd hook to lock the computer if I were to suspend it on purpose. I followed the arch wiki for this and found this link: https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#Sleep_hooks.

While trying to enable and run the service, hyprlock is core dumping:
```
× user-suspend@user.service - User suspend actions - Lock screen
     Loaded: loaded (/etc/systemd/system/user-suspend@.service; enabled; preset: disabled)
     Active: failed (Result: core-dump) since Tue 2024-07-02 23:14:56 AEST; 59s ago
 Invocation: a625f5aa5e994c0aa7d582e950871745
    Process: 70489 ExecStart=/usr/bin/hyprlock (code=dumped, signal=SEGV)

Jul 02 23:14:55 eos-laptop systemd[1]: Starting User suspend actions - Lock screen...
Jul 02 23:14:56 eos-laptop systemd[1]: user-suspend@user.service: Control process exited, code=dumped, status=11/SEGV
Jul 02 23:14:56 eos-laptop systemd[1]: user-suspend@user.service: Failed with result 'core-dump'.
Jul 02 23:14:56 eos-laptop systemd[1]: Failed to start User suspend actions - Lock screen.

```

I had receive some other errors earlier (but strangely didn't receive them this time) when I tried the same:
```
Jul 02 22:59:46 eos-laptop hyprlock[50039]: error: XDG_RUNTIME_DIR is invalid or not set in the environment.
Jul 02 22:59:46 eos-laptop hyprlock[50039]: [CRITICAL] Couldn't connect to a wayland compositor
```

I was able to run hyprlock without any issues on the as a systemd user service but it was pointless since the systemd user services can't get the sleep.target. So, the highly inexperience me is thinking that this has to do with running hyprlock as a root or something.

PaideiaDilemma:
Do we prefer throwing an error like this, or logging + exit or just returning an empty string?

Same problem in hyprland as well.

Closes #392

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -23,7 +23,12 @@ static std::string getConfigDir() {
     if (xdgConfigHome && std::filesystem::path(xdgConfigHome).is_absolute())
         return xdgConfigHome;
 
-    return getenv("HOME") + std::string("/.config");
+    static const char* home = getenv("HOME");
+
+    if (!home)
+        throw std::runtime_error("Neither HOME nor XDG_CONFIG_HOME is set in the environment. Cannot determine config directory.");
+
+    return home + std::string("/.config");
 }
 
 static std::string getMainConfigPath() {


vaxerski (approved):
top g

PR #399, opened 2024-07-05
auth: pam fallback log message and always call pam_end after pam_authenticate

PaideiaDilemma:
Changed the log, because someone used an absolute path for `general:pam_module`.

Also new release maybe?
People that use 0.3.0 are quite a bit behind by now and don't have d9f644125468054d2a237adf4ae676f64c0da918.

### DIFF
--- a/src/core/Auth.cpp
+++ b/src/core/Auth.cpp
@@ -91,6 +91,8 @@ bool CAuth::auth() {
     }
 
     ret = pam_authenticate(handle, 0);
+    pam_end(handle, ret);
+    handle = nullptr;
 
     m_sConversationState.waitingForPamAuth = false;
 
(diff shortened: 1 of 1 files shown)


src/core/Auth.cpp
```diff
@@ -91,6 +91,8 @@ bool CAuth::auth() {
     }
 
     ret = pam_authenticate(handle, 0);
+    pam_end(handle, ret);
+    handle = NULL;
```
vaxerski, line 95: nullptr

vaxerski:
> Also new release maybe?

sure.

src/core/Auth.cpp
PaideiaDilemma: thx

vaxerski (approved):
thanks

PR #405, opened 2024-07-07
"Some bug fixing"

JassonCordones:
Fixes some bugs in the LockSurface

- Surface Creation Check in Constructor: Replaced !surface check with !lockSurface to correctly handle creation failure
- Surface Destruction in Destructor: Added a check if (surface) before wl_surface_destroy(surface) 
- Verify that fractionalScale calculation is correct.
- Improved error handling around eglSurface creation.
- Proper synchronization and cleanup of frameCallback: Set needsFrame to false on Callback before calling render().


vaxerski (approved):
thansk!

PR #406, opened 2024-07-07
config: add input-field fail_timeout option

davc0n:


src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -328,6 +334,7 @@ void CPasswordInputField::updatePlaceholder() {
     placeholder.asset = nullptr;
 
     if (placeholder.isFailText) {
+        g_pHyprlock->addTimer(std::chrono::seconds(configFailTimeoutSec), failTimeoutCallback, nullptr);
```
PaideiaDilemma, line 337: Should be milliseconds, like `fade_timeout` I think

src/renderer/widgets/PasswordInputField.cpp
```diff
@@ -304,6 +305,11 @@ bool CPasswordInputField::draw(const SRenderData& data) {
     return dots.currentAmount != passwordLength || fade.animated || col.animated || redrawShadow || data.opacity < 1.0 || forceReload;
 }
 
+static void failTimeoutCallback(std::shared_ptr<CTimer> self, void* data) {
+    g_pAuth->m_bDisplayFailText = false;
+    g_pHyprlock->renderAllOutputs();
+}
+
```
PaideiaDilemma, line 312: So the thing about moving this to `PasswordInputField` is that it will be called for each monitor separately. So you would have one timer per monitor. Not that big of a deal, but i think it is better to just have it in the `onPasswordCheckTimer` function to avoid that.

davc0n: Tbh it has been moved just to be consistent with the configuration parameters which is input-field->fail_timeout.
Not sure what's the proper way to access that option outside `CPasswordInputField`.

PaideiaDilemma: Yeah makes sense, since you could have multiple input fields theoretically. I think it is fine link this.

src/renderer/widgets/PasswordInputField.hpp
```diff
@@ -39,6 +39,7 @@ class CPasswordInputField : public IWidget {
     Vector2D    configSize;
 
     std::string halign, valign, configFailText, outputStringPort, configPlaceholderText;
+    u_int64_t   configFailTimeoutMs;
```
vaxerski, line 42: `uint64_t`

also please initialize it.

davc0n: Typo, sorry.
Is `uint64_t    configFailTimeoutMs = 0;` correct, or it should be 2000? I though the following line was enough:

```
configFailTimeoutMs      = std::any_cast<Hyprlang::INT>(props.at("fail_timeout"));
```

There is `fadeTimeoutMs` at the end of the class which is not initialized either, should that be fixed too?

vaxerski (approved):
thanks! wiki mr needed

PR #407, opened 2024-07-07
core: add option to render solid background immediatly when bg assets are not ready

### LINKED ISSUE #389: Feature request: Immediately black out the screen when `--immediate` flag is used
When locking the screen before the PC suspends, it's important that this happens *fast*. It's really disconcerting when the monitor is visible for a couple of seconds after waking the PC before the lock screen fades in, and it could conceivably be a security issue in some cases.

I think it would be nice if the `--immediate` flag caused hyprlock to black out the screen as soon as possible, before doing any other non-critical setup like loading images.

This would ensure that when the machine resumes, there's only a black screen while hyprlock finishes its setup and can fade in the nice lock background image.

For backwards compatibility this could also be a new flag with the name `--sleep`, `--blackout` etc.

PaideiaDilemma:
Closes #389

Option and flag that allows `background:color` to be rendered immediately, when the background assets are not ready yet.


vaxerski (approved):
thansk!

PR #410, opened 2024-07-07
Some improvements to log & FrameBuffer + misc warning fixes

JassonCordones:
**Log**
- Added separate function for formatting and logging.
- Used std::abort() to terminate the program instead of de-referencing null pointer.

**FrameBuffer**
-Added check to ensure m_pStencilTex is valid before usage.

**Other files**
-Misc warning fixes and formatting.


src/helpers/Log.hpp
```diff
@@ -28,6 +28,11 @@ namespace Debug {
     inline bool quiet   = false;
     inline bool verbose = false;
 
+    template <typename... Args>
+    std::string format(const std::string& fmt, Args&&... args) {
```
vaxerski, line 32: what for?

PR #413, opened 2024-07-08
Label exception handling + Frambuffer checks + headers

JassonCordones:
**Label**
- Added exception handling in label constructor
**Log**
- Used std::abort() to terminate the program instead of de-referencing null pointer.
**Frambuffer**
- Added check to ensure m_pStencilTex is valid before usage.
-  sets m_pStencilTex to nullptr after release.
**Misc**
- Headers


vaxerski (approved):
thanks!

PR #414, opened 2024-07-10
asyncResourceGatherer: trace logs and exit behaviour

PaideiaDilemma:
Trace logs should help debug label update problems.

`dmas.clear` in the asyncLoopThread sometimes caused a crash. So i removed it. Will get destroyed together with the asyncResourceGatherer. Also tought clearing requests would be a good idea when calling `AsyncResourceGatherer::notify`.


vaxerski (approved):
thanks!

PR #418, opened 2024-07-11
core: handle ext_session_lock_v1::finished as defined by the protocol

### LINKED ISSUE #416: [Weird Behaviour]: multiple hyprlock istances keep running
If i run hyprlock, and then once it's running, i run an other instance of hyprlock (through a keybind with bindl, or on switch lid for example), then the first istance gets closed after i unlock the screen, but the other istances keep running in the background, without doing anything (ie, if i killall hyprlock, i don't get the red screen of death)

I can think of two ways of fixing this:
- not running hyprlock, if an other istance is already running
- kill all hyprlock instances running in the background once the screen is unlocked

But in both cases, someone may want to have multiple istances of hyprlock running for multiple displays. 

So, since solutions like keeping track of if  hyprlock is running for every display seems a very big complication for me,  maybe adding a flag like `--single-instance` could make sense instead?

PaideiaDilemma:
Closes #416

According to [ext_session_lock_v1::finished](https://wayland.app/protocols/ext-session-lock-v1#ext_session_lock_v1:event:finished)

>  Upon receiving this event, the client should make either the destroy request or the unlock_and_destroy request, depending on whether or not the locked event was received on this object.

Also `exit` previously did not terminate hyprlock properly. I think the right thing to do is to set `m_bTerminate`.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -959,8 +959,15 @@ void CHyprlock::onLockLocked() {
 
 void CHyprlock::onLockFinished() {
     Debug::log(LOG, "onLockFinished called. Seems we got yeeten. Is another lockscreen running?");
-    g_pRenderer = nullptr;
-    exit(1);
+    if (m_bLocked)
+        // The `finished` event specifies that whenever the `locked` event has been recieved and the compositor sends `finished`,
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thanks!

PR #421, opened 2024-07-14
core: immediately create session lock surfaces

### LINKED ISSUE #420: Red screen before Hyprlock starts
When starting hyprlock on latest hl `main`, it first shows a red screen for 1-2s, and only then shows hyprlock itself. This also causes hyprlock's screenshot (used for blurring, etc.) to be solid red.

I believe that this is a regression caused by https://github.com/hyprwm/Hyprland/pull/6850 on hyprland's side. I've tested other lockscreens like swaylock and gtklock, which work correctly the mentioned PR.

Here's a video how it looks on my end (~~oh no you now know my passwort length, because ffmpeg doesn't want to crop for some reason~~):

https://github.com/user-attachments/assets/e2d2060a-9805-4f46-8ab9-670e38e262f6

PaideiaDilemma:
Instead of waiting for the `locked` event, create session lock surfaces right away.

Closes #420

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -400,6 +400,9 @@ void CHyprlock::run() {
 
     acquireSessionLock();
 
+    if (m_bTerminate) // Recieved finished
+        exit(1);
+
     g_pAuth = std::make_unique<CAuth>();
     g_pAuth->start();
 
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks!

PR #424, opened 2024-07-17
core: grace unlock improvements and auth fixes for grace/SIGUSR1 unlocks

### LINKED ISSUE #423: Grace cause fail auth
When set grace>0 and unlock by it

It trigger several times
```
[LOG] In grace and cursor moved more than 5px, unlocking!
[LOG] Unlocking session
[ERR] Unlock already happend?
```
and its be count in faillock times

PaideiaDilemma:
Closes #423

Grace code previously respected the case when `m_bFadeStarted` was already set, but not `m_bTerminate` in case `gerneral:no_fade_out` is used.

Auth needs to check for `isUnlocked` to not cause a failed authentication attempt. Also we don't need to enqueue `passwordCheckTimerCallback` if we are already unlocked.

I didn't like `success` being part of `m_sConversationState`. So I moved it outside of it, used the return value of `auth` to set it and renamed it to `m_bAuthenticated`.

https://github.com/hyprwm/hyprlock/pull/424/commits/b6d2be62184dc3e06ed73f16d1ef4211d1433617 is just in there because i noticed it.

### DIFF
--- a/src/core/Auth.cpp
+++ b/src/core/Auth.cpp
@@ -38,7 +38,7 @@ int conv(int num_msg, const struct pam_message** msg, struct pam_response** resp
                 }
 
                 // Needed for unlocks via SIGUSR1
-                if (g_pHyprlock->m_bTerminate)
+                if (g_pHyprlock->isUnlocked())
                     return PAM_CONV_ERR;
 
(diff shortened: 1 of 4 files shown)


vaxerski (approved):
thanks!

Issue #390, opened 2024-07-01
`$TIME` label doesn't update


ThinkChaos:
When leaving hyprlock running for a while, `$TIME` stops updating. It seems to stop a couple minutes after locking, and re-updates itself a bit after user input.
I noticed that a `cmd` label does update, so with the config below, the cmd date is on time while `$TIME` stops updating. So it seems to be `$TIME` specific (or maybe any label without `cmd[update]`).

I was hoping this would be resolved at the same time as #357, but I still have the issue even with 318c00d6d093d6d0da2687ad51880578a49899dd.
Just in case, I have dual screens and an AMD GPU.

<details>

<summary>config</summary>

```
label {
  monitor=DP-1
  font_family=monospace
  font_size=150
  text=$TIME
  text_align=center
}

label {
  monitor=DP-1
  font_family=monospace
  font_size=150
  halign=center
  text=cmd[update:1000] date
  text_align=center
  valign=center
}
```
</details>

PaideiaDilemma:
Hmm can you check if `Trying to update label, but resource {} is still pending! Skipping update.` is in the logs when it happens?
I will see if I can reproduce it.

PaideiaDilemma:
Does not happen on my side. Also AMD GPU and multi monitor. I think the fact that `$TIME` is used does not matter. To be honest i don't understand how that can happen after https://github.com/hyprwm/hyprlock/commit/318c00d6d093d6d0da2687ad51880578a49899dd

ThinkChaos:
Indeed that log appears. Here's my service's log, with artificial paragraphs to highlight the pattern.
Note that this was generated from a slightly different config than posted above since it's from my actual setup.

```
Jul 03 18:09:00 Workstation hyprlock[1537280]: [LOG] Configuring surface for logical [Vector2D: x: 3840, y: 2160] and pixel [Vector2D: x: 3840, y: 2160]
Jul 03 18:09:00 Workstation hyprlock[1537280]: [ERR] Unknown prop in string format
Jul 03 18:09:00 Workstation hyprlock[1537280]: [LOG] got fractional 1
Jul 03 18:09:00 Workstation hyprlock[1537280]: [LOG] configure with serial 176420
Jul 03 18:09:00 Workstation hyprlock[1537280]: [LOG] Configuring surface for logical [Vector2D: x: 1920, y: 1080] and pixel [Vector2D: x: 1920, y: 1080]
[...]

PR #433, opened 2024-07-19
asyncResourceGatherer: deduplicate image rendering code

PaideiaDilemma:
Also makes the image widget support JPEG and WEBP.

Tested all three file types and backgrounds still render the same.

### DIFF
--- a/src/renderer/AsyncResourceGatherer.cpp
+++ b/src/renderer/AsyncResourceGatherer.cpp
@@ -91,7 +91,7 @@ enum class FileType {
     UNKNOWN,
 };
 
-FileType getFileType(std::filesystem::path path) {
+FileType getFileType(const std::filesystem::path& path) {
     std::string ext = path.extension().string();
     // convert the extension to lower case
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
generally I want to move the image loading (+ the async resource gatherer) to a new library at some point in the future, to de-dup between here, hyprpaper, and future stuff

src/renderer/widgets/Image.cpp
```diff
@@ -190,14 +190,6 @@ bool CImage::draw(const SRenderData& data) {
     return data.opacity < 1.0;
 }
 
-static void onAssetCallbackTimer(std::shared_ptr<CTimer> self, void* data) {
```
vaxerski, line 193: why is this gone

PaideiaDilemma: Cause i also removed that from `Label` in #384
I thought we don't need it any more, since we set `needsFrame = true`, when `frameCallback` is present in `CSessionLockSurface::render`.

But maybe readd it just to be safe and maybe fix #390?

vaxerski: I just dont think it belongs in this mr

PaideiaDilemma: Fair

vaxerski (approved):
thanks!

PR #440, opened 2024-07-24
misc: add issue templates

PaideiaDilemma:
Copied from Hyprland and slightly modified.

I think this will reduce the amount of times we have to ask for versions of hyprlock and hyprland.

~_I don't know if it is enough to just merge this to set the templates..._~

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,102 @@
+name: Bug Report
+description: Something is not working right
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Before opening a new issue, please take a moment to search through the current open and closed issues to check if it already exists.
+
+        ---
+
+  - type: dropdown
+    id: type
+    attributes:
+      label: Regression?
+      description: |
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,102 @@
...
+      required: true
+
+  - type: textarea
+    id: ver
+    attributes:
+      label: Hyprlock Info and Version
+      description: |
+        Please check and enter your hyprlock version.
+        If you are on a packaged version, use your package manager to find your hyprlock version.
+        If you built it yourself, please provide the commit hash.
```
PaideiaDilemma, line 33: We might want to build `--version` into hyprlock.
Is that something that would be nice to have in hyprutils?

PaideiaDilemma: Nevermind, cannot be in hyprutils, since it is a build time feature right?

vaxerski: yea, would be nice.

vaxerski: done

vaxerski (approved):
coolio

Issue #452, opened 2024-08-03
Command line flag to one-shot disable fade in


alba4k:
### Description

It would be great to have a flag to disable fade-in, just like there is for disabling the grace period, without having to create a whole separate config just for that. Something like `--immediate`, maybe `--no-fadein`

PR #453, opened 2024-08-03
renderer: add `--no-fade-in`

alba4k:
This PR implements what I wished for in https://github.com/hyprwm/hyprlock/issues/452

It simply adds a `--no-fade-in` flag that disables the fade in animation when used, overwriting the use of no_fade_in (as if it was always set to true).

For the implementation, I mainly looked at the way `--immediate-render` was implemented.

LMK if there are any changes I should make, the main doubts I have are:
* is the internal `m_bNoFafeIn` name correct? not sure what the m and the b stand for.
* the commit name, should it start with `renderer:`? Would something else like `core:` be better?

### DIFF
--- a/src/core/hyprlock.hpp
+++ b/src/core/hyprlock.hpp
@@ -25,7 +25,7 @@ struct SDMABUFModifier {
 
 class CHyprlock {
   public:
-    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender);
+    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const bool noFadeIn);
     ~CHyprlock();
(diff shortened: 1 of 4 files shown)


src/core/hyprlock.hpp
```diff
@@ -102,6 +102,9 @@ class CHyprlock {
     bool                            m_bFadeStarted = false;
 
     bool                            m_bImmediateRender = false;
+
+    bool                            m_bNoFadeIn = false;
```
alba4k, line 106: is this name correct?

vaxerski: real

vaxerski (approved):

Issue #65, opened 2024-02-22
[bug] When switching monitors getting red screen


hrasekj:
Hello,

on my laptop, when i have locked screen and change the monitor, the hyprlock ends up in some error state with red screen.

I've tried different configurations and looks it has no effect. 

Also i tried to call pkill from different session and ends up with the red screen to :thinking: 

Here is `coredumpctl debug hyprlock`
```
           PID: 15258 (hyprlock)
           UID: 1000 (hrasekj)
           GID: 1000 (hrasekj)
        Signal: 11 (SEGV)
     Timestamp: Thu 2024-02-22 21:21:43 CET (11min ago)
  Command Line: hyprlock
    Executable: /usr/local/bin/hyprlock
 Control Group: /user.slice/user-1000.slice/session-3.scope
          Unit: session-3.scope
         Slice: user-1000.slice
       Session: 3
     Owner UID: 1000 (hrasekj)
       Boot ID: 65d1cd55441e4864901e98f8eee23454
    Machine ID: f8364d72622840a0ae35bb54cb91a894
      Hostname: byte-side
       Storage: /var/lib/systemd/coredump/core.hyprlock.1000.65d1cd55441e4864901e98f8eee23454.15258.1708633303000000.zst (present)
  Size on Disk: 2.2M
       Message: Process 15258 (hyprlock) of user 1000 dumped core.
                
                Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2.amd64
                Stack trace of thread 15258:
                #0  0x00007f47c256767b n/a (libwayland-client.so.0 + 0x967b)
                #1  0x00007f47c2567ddd n/a (libwayland-client.so.0 + 0x9ddd)
                #2  0x00007f47c256540a wl_proxy_marshal_flags (libwayland-client.so.0 + 0x740a)
                #3  0x0000557770e099d8 _ZN19CSessionLockSurface6renderEv (hyprlock + 0x299d8)
                #4  0x0000557770e300bf _ZL7onTimerSt10shared_ptrI6CTimerEPv (hyprlock + 0x500bf)
                #5  0x0000557770e0daa2 _ZNSt17_Function_handlerIFvSt10shared_ptrI6CTimerEPvEPS4_E9_M_invokeERKSt9_Any_dataOS2_OS3_ (hyprlock + 0x2daa2)
                #6  0x0000557770e0dc86 _ZN6CTimer4callESt10shared_ptrIS_E (hyprlock + 0x2dc86)
[...]

PR #462, opened 2024-08-17
lockSurface: fix dynamic output mode and scale updates

PaideiaDilemma:
Fixes dynamic output updates via programs like kanshi or wdisplays. (Ref #65)

Accidentally fixed fractional scaling on river and sway (hence e34f67a4f9a861bd05fb5a4a237b4539ea45bf70).
Turns out wlroots sends a correct configure event at some point, but we already created widgets with a wrong viewport. So dynamically removing them fixes it.

Now instead of destroying all widgets for the surface, we could also just update the viewport of all widgets.
Did not do that yet, since I did not want to give the IWidget base class a state and did not feel like implementing a viewport setter for all widgets. Let me know which one of the two sounds better, or if removing like this is fine (widgets will just get recreated with `getOrCreatedWidgetsFor`).


vaxerski (approved):
jaaager

PR #464, opened 2024-08-20
fix crash caused by exiting without joining running thread

yangyingchao:
Hyprlock crashes frequently in my case. When a crash occurs, GDB shows:

1. The main thread is trying to exit, and `CAsyncResourceGatherer` has already been destroyed.
2. `initialGatherThread` is still running, walking through `dmas` of the destroyed `CAsyncResourceGatherer` instance.

Calling await for `initialGatherThread` should fix this specific crash.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -405,8 +405,14 @@ void CHyprlock::run() {
 
     acquireSessionLock();
 
-    if (m_bTerminate) // Recieved finished
+    // Recieved finished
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
clang-format pls

yangyingchao:
It's been formatted already. The bracket was placed on a new line due to the trailing comment, I guess.

I moved the comment to a separate line so clang-format can place the bracket at the end of the if statement.

vaxerski (approved):

PR #489, opened 2024-09-15
Added the option to daemonize the process

stojanov:
Added the option for the process to be forked when starting.

Why would this be useful ?
I tried running hyprlock from wlogout and then suspending the system, but that didn't work since the system would first lock and then when unlocking would suspend, this should solve that problem

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4,6 +4,7 @@
 #include "src/helpers/Log.hpp"
 #include <cstddef>
 #include <iostream>
+#include <unistd.h>
 
 void help() {
     std::cout << "Usage: hyprlock [options]\n\n"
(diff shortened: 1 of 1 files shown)


vaxerski:
what

stojanov:
Ahh sorry, misstyped, after doing `hyprlock && systemctl suspend` what would happen is first the system would lock and after that unlocking/exiting hyprlock the system woud suspend immediately. This is gust forking the process and letting the suspend command run

vaxerski:
what's the point? Just run `hyprlock & systemctl suspend`?

stojanov:
Well, in a sense, yes. I'm not sure what else this will be useful in. I'm not sure if this also aligns with the idea of hyprlock.

stojanov:
Ahh i see, that is also useful. Sorry for my useless implementation

PR #495, opened 2024-09-23
Documentation mtx.hpp + minor readability fixes

JassonCordones:
Title


vaxerski:
mtx changes can be dropped. Entirety of mtx will be rewritten by me soon™ into hyprutils

JassonCordones:
That would be basically dropping the whole PR since the changes to other files are mainly name changes where mtx functions are used.

vaxerski:
then that's what should be done tbh

PR #510, opened 2024-10-11
input-field: add dots_text_format to support setting arbitrary chars as the input indicator

### LINKED ISSUE #316: [Feature Request] render password as text symbol
image
- image is from : https://github.com/stepanzubkov/where-is-my-sddm-theme/commits?author=stepanzubkov

I want to render password dots as asterisk like this image.

I looked up code little bit and find out dots are rendered by CRenderer::renderRect function. Is there any plan to render password as text?

### LINKED ISSUE #482: Custom input-field character?
### Description

I am trying to copy my [SDDM theme](https://github.com/stepanzubkov/where-is-my-sddm-theme) and would like if there was an option to change the input-field character to an asterisk (*) instead of the current dot.

PaideiaDilemma:
This adds an option, that allows setting an arbitrary char to be rendered instead of a rectangle for the input fields dots.

Closes #316
Closes #482

Let me know if you think we should name the option something else.
Guy in #316 suggested `dots_mask_character`. Maybe just `dots_char`?
Currently the option does support multiple chars as well, so I chose `dots_text`. But in that case `dots_text_format` is a bit more clear I think.

I also added the `font_familiy` option to the input-field, in order for users to be able to use a nerd font for example.

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -110,6 +110,7 @@ void CConfigManager::init() {
     m_config.addSpecialConfigValue("input-field", "fade_on_empty", Hyprlang::INT{1});
     m_config.addSpecialConfigValue("input-field", "fade_timeout", Hyprlang::INT{2000});
     m_config.addSpecialConfigValue("input-field", "font_color", Hyprlang::INT{0xFF000000});
+    m_config.addSpecialConfigValue("input-field", "font_family", Hyprlang::STRING{"Sans"});
     m_config.addSpecialConfigValue("input-field", "halign", Hyprlang::STRING{"center"});
     m_config.addSpecialConfigValue("input-field", "valign", Hyprlang::STRING{"center"});
     m_config.addSpecialConfigValue("input-field", "position", Hyprlang::VEC2{0, -20});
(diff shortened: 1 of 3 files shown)


PaideiaDilemma:
One thing that I want to mention is that the size and spacing between dots depends on the font and character.

Here are some samples:
For all of them `dots_spacing=0.2` and `dots_size=0.5`

    dots_text_format=*
    # default
    font_family=Sans
    
sans_mult

    dots_text_format=🔥
    font_family=Sans
   
**Notice the increased spacing. Can be adjusted with dots_spacing**   

sans_with_emoji


    dots_text_format=
    font_family=FiraCode Nerd Font

nerd_font_hl_symbol



    dots_text_format=
    font_family=FiraCode Nerd Font Mono


nerd_font_mono_symbol

</details>

PaideiaDilemma:
I just noticed, that I cannot set `dots_spacing` to be negative, which could be useful for emojis for example. But it should work with a `Hyprlang::FLOAT` right?

vaxerski:
yeah, it should. Hyprlang-test seems to work with negative floats. Make sure you don't clamp it anywhere in hyprlock

PaideiaDilemma:
yeah thanks! I missed that it was clamped. It is now allowed from -1 to 1.

vaxerski:
oopsie indeed

vaxerski (approved):

vaxerski:
where is the wiki mr

PaideiaDilemma:
[Here](https://github.com/hyprwm/hyprland-wiki/pull/805), just wanted to wait if someone has an opinion on the options name.

PR #512, opened 2024-10-11
misc: use hyprutils for string replaceAll

PaideiaDilemma:


PaideiaDilemma:
Next we could rip out `mtx.hpp` and replace it with Mat3x3 from hyprutils right?

vaxerski:
ya

vaxerski (approved):

PR #514, opened 2024-10-11
auth: Support parallel fingerprint auth

### LINKED ISSUE #258: [Feature] Support parallel unlocking with fingerprint and password
The only thing I've seen that supports this correctly so far is GDM, which runs 2 PAM sessions in parallel, one for fingerprint, one for password. This makes the whole "press enter to trigger the prompt" song and dance unnecessary.

I'm unsure of how this actually works in terms of implementation, but only that it is possible (since GDM can do it).

Would be very nice if Hyprlock could support it.

moggiesir:
I chose to use Fprint's dbus interface directly rather than going through pam (which uses Fprint's dbus interface) due to poor handling of system sleep somewhere between fprintd and pam. When preparing for sleep, fprintd puts the device to sleep, which causes VerifyStatus to emit with verify-unknown-error, which normally should be responded to by calling both Device.StopVerify and Device.Release (and this is what pam does). Unfortunately, if you try to release the device when the system is preparing for sleep, you'll get an error that the device is busy and then you can't can't claim or release the device for 30 seconds.

pam also has a max timeout for pam_fprintd.so of 99 seconds, and so if we used pam, we'd have to deal with the timeouts and keep restarting the auth conversation.

gdm/gnome-session lock seems to get around these issues by having a shutter on top of the lock screen that you have to interact with first that gives gnome-session a trigger to start fingerprint auth.

Fixes #258

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,8 @@ pkg_check_modules(
   pangocairo
   libdrm
   gbm
-  hyprutils>=0.2.3)
+  hyprutils>=0.2.3
+  sdbus-c++>=2.0.0)
 
 file(GLOB_RECURSE SRCFILES CONFIGURE_DEPENDS "src/*.cpp")
 add_executable(hyprlock ${SRCFILES})
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -50,6 +50,9 @@ void CConfigManager::init() {
     m_config.addConfigValue("general:immediate_render", Hyprlang::INT{0});
     m_config.addConfigValue("general:pam_module", Hyprlang::STRING{"hyprlock"});
     m_config.addConfigValue("general:fractional_scaling", Hyprlang::INT{2});
+    m_config.addConfigValue("general:enable_fingerprint", Hyprlang::INT{0});
+    m_config.addConfigValue("general:fingerprint_ready_message", Hyprlang::STRING{"(Scan fingerprint to unlock)"});
+    m_config.addConfigValue("general:fingerprint_present_message", Hyprlang::STRING{"Scanning fingerprint"});
 
     m_config.addSpecialCategory("background", Hyprlang::SSpecialCategoryOptions{.key = nullptr, .anonymousKeyBased = true});
     m_config.addSpecialConfigValue("background", "monitor", Hyprlang::STRING{""});
--- a/src/core/Fingerprint.cpp
+++ b/src/core/Fingerprint.cpp
@@ -1,0 +1,240 @@
+#include "Fingerprint.hpp"
+#include "../helpers/Log.hpp"
+#include "../config/ConfigManager.hpp"
+
+#include <filesystem>
+#include <unistd.h>
+#include <pwd.h>
+
+#include <cstring>
+#include <thread>
+
+static const auto FPRINT        = sdbus::ServiceName{"net.reactivated.Fprint"};
+static const auto DEVICE        = sdbus::ServiceName{"net.reactivated.Fprint.Device"};
+static const auto MANAGER       = sdbus::ServiceName{"net.reactivated.Fprint.Manager"};
+static const auto LOGIN_MANAGER = sdbus::ServiceName{"org.freedesktop.login1.Manager"};
+static const auto RETRY_MESSAGE = "Could not match fingerprint. Try again.";
+
+enum MatchResult {
+    MATCH_INVALID = 0,
+    MATCH_NO_MATCH,
+    MATCH_MATCHED,
+    MATCH_RETRY,
+    MATCH_SWIPE_TOO_SHORT,
+    MATCH_FINGER_NOT_CENTERED,
+    MATCH_REMOVE_AND_RETRY,
+    MATCH_DISCONNECTED,
+    MATCH_UNKNOWN_ERROR,
+};
+
+static std::map<std::string, MatchResult> s_mapStringToTestType = {{"verify-no-match", MATCH_NO_MATCH},
+                                                                   {"verify-match", MATCH_MATCHED},
+                                                                   {"verify-retry-scan", MATCH_RETRY},
+                                                                   {"verify-swipe-too-short", MATCH_SWIPE_TOO_SHORT},
+                                                                   {"verify-finger-not-centered", MATCH_FINGER_NOT_CENTERED},
+                                                                   {"verify-remove-and-retry", MATCH_REMOVE_AND_RETRY},
+                                                                   {"verify-disconnected", MATCH_DISCONNECTED},
+                                                                   {"verify-unknown-error", MATCH_UNKNOWN_ERROR}};
+
+CFingerprint::CFingerprint() {
+    static auto* const PFINGERPRINTREADY   = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:fingerprint_ready_message"));
+    m_sFingerprintReady                    = *PFINGERPRINTREADY;
+    static auto* const PFINGERPRINTPRESENT = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:fingerprint_present_message"));
+    m_sFingerprintPresent                  = *PFINGERPRINTPRESENT;
+    static auto* const PENABLEFINGERPRINT  = (Hyprlang::INT* const*)g_pConfigManager->getValuePtr("general:enable_fingerprint");
+    m_bEnabled                             = **PENABLEFINGERPRINT;
+}
+
+std::shared_ptr<sdbus::IConnection> CFingerprint::start() {
+    if (!m_bEnabled)
+        return {};
+    m_sDBUSState.connection = sdbus::createSystemBusConnection();
+    registerSleepHandler();
+
+    // When entering sleep, the wake signal will trigger startVerify().
+    if (m_sDBUSState.sleeping)
+        return m_sDBUSState.connection;
+
+    startVerify();
+
+    return m_sDBUSState.connection;
+}
+
+bool CFingerprint::isAuthenticated() {
+    return m_bAuthenticated;
+}
+
+std::optional<std::string> CFingerprint::getLastMessage() {
+    return m_sDBUSState.message.empty() ? std::nullopt : std::optional(m_sDBUSState.message);
+}
+
+void CFingerprint::terminate() {
+    if (!m_sDBUSState.abort)
+        releaseDevice();
+}
+
+void CFingerprint::registerSleepHandler() {
+    m_sDBUSState.login    = sdbus::createProxy(*m_sDBUSState.connection, sdbus::ServiceName{"org.freedesktop.login1"}, sdbus::ObjectPath{"/org/freedesktop/login1"});
+    m_sDBUSState.sleeping = m_sDBUSState.login->getProperty("PreparingForSleep").onInterface(LOGIN_MANAGER).get<bool>();
+    m_sDBUSState.login->uponSignal("PrepareForSleep").onInterface(LOGIN_MANAGER).call([this](bool start) {
+        Debug::log(LOG, "fprint: PrepareForSleep (start: {})", start);
+        if (start) {
+            m_sDBUSState.sleeping = true;
+            stopVerify();
+            m_sDBUSState.inhibitLock.reset();
+        } else {
+            m_sDBUSState.sleeping = false;
+            inhibitSleep();
+            startVerify();
+        }
+    });
+    if (!m_sDBUSState.sleeping)
+        inhibitSleep();
+}
+
+void CFingerprint::inhibitSleep() {
+    m_sDBUSState.login->callMethod("Inhibit")
+        .onInterface(LOGIN_MANAGER)
+        .withArguments("sleep", "hyprlock", "Fingerprint verifcation must be stopped before sleep", "delay")
+        .storeResultsTo(m_sDBUSState.inhibitLock);
+}
+
+bool CFingerprint::createDeviceProxy() {
+    auto              proxy = sdbus::createProxy(*m_sDBUSState.connection, FPRINT, sdbus::ObjectPath{"/net/reactivated/Fprint/Manager"});
+
+    sdbus::ObjectPath path;
+    try {
+        proxy->callMethod("GetDefaultDevice").onInterface(MANAGER).storeResultsTo(path);
+    } catch (sdbus::Error& e) {
+        Debug::log(WARN, "fprint: couldn't connect to Fprint service ({})", e.what());
+        return false;
+    }
+    Debug::log(LOG, "fprint: using device path {}", path.c_str());
+    m_sDBUSState.device = sdbus::createProxy(*m_sDBUSState.connection, FPRINT, path);
+
+    m_sDBUSState.device->uponSignal("VerifyFingerSelected").onInterface(DEVICE).call([](const std::string& finger) { Debug::log(LOG, "fprint: finger selected: {}", finger); });
+    m_sDBUSState.device->uponSignal("VerifyStatus").onInterface(DEVICE).call([this](const std::string& result, const bool done) { handleVerifyStatus(result, done); });
+
+    m_sDBUSState.device->uponSignal("PropertiesChanged")
+        .onInterface("org.freedesktop.DBus.Properties")
+        .call([this](const std::string& interface, const std::map<std::string, sdbus::Variant>& properties) {
+            if (interface != DEVICE || m_sDBUSState.done)
+                return;
+
+            try {
+                const auto presentVariant = properties.at("finger-present");
+                bool       isPresent      = presentVariant.get<bool>();
+                if (!isPresent)
+                    return;
+                m_sDBUSState.message = m_sFingerprintPresent;
+                g_pHyprlock->enqueueForceUpdateTimers();
+            } catch (std::out_of_range& e) {}
+        });
+
+    return true;
+}
+
+void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
+    Debug::log(LOG, "fprint: handling status {}", result);
+    auto matchResult = s_mapStringToTestType[result];
+    if (m_sDBUSState.sleeping && matchResult != MATCH_DISCONNECTED)
+        return;
+    switch (matchResult) {
+        case MATCH_INVALID: Debug::log(WARN, "fprint: unknown status: {}", result); break;
+        case MATCH_NO_MATCH:
+            stopVerify();
+            if (m_sDBUSState.retries >= 3) {
+                m_sDBUSState.message = "Fingerprint auth disabled: too many failed attempts";
+            } else if (startVerify(/* showMessage= */ false)) {
+                done = false;
+                m_sDBUSState.retries++;
+                m_sDBUSState.message = RETRY_MESSAGE;
+            } else {
+                m_sDBUSState.message = "Fingerprint auth disabled: could not restart verification";
+            }
+            break;
+        case MATCH_UNKNOWN_ERROR:
+            stopVerify();
+            m_sDBUSState.message = "Unknown fingerprint error, disabling fingerprint auth";
+            break;
+        case MATCH_MATCHED:
+            stopVerify();
+            m_bAuthenticated     = true;
+            m_sDBUSState.message = "";
+            g_pHyprlock->unlock();
+            break;
+        case MATCH_RETRY: m_sDBUSState.message = "Please retry fingerprint scan"; break;
+        case MATCH_SWIPE_TOO_SHORT: m_sDBUSState.message = "Swipe too short - try again"; break;
+        case MATCH_FINGER_NOT_CENTERED: m_sDBUSState.message = "Finger not centered - try again"; break;
+        case MATCH_REMOVE_AND_RETRY: m_sDBUSState.message = "Remove your finger and try again"; break;
+        case MATCH_DISCONNECTED:
+            m_sDBUSState.message = "Fingerprint device disconnected";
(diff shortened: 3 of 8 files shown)


src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,237 @@
...
+        case MatchResult::NoMatch: //The verification did not match, Device.VerifyStop should now be called.
+            stopVerify();
+            if (m_sDBUSState.retries < 3) {
+                startVerify();
+                done = false;
+                m_sDBUSState.retries++;
+                m_sDBUSState.failText = RETRY_MESSAGE;
+                if (clearRetryTextTimer) clearRetryTextTimer->cancel();
+                clearRetryTextTimer   = g_pHyprlock->addTimer(
+                    std::chrono::seconds(2),
```
PaideiaDilemma, line 146: This could just be the `input-field:fail_timeout` option I think?

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return m_bAuthenticated;
+}
+
+std::optional<std::string> CFingerprint::getLastFailText() {
+    return m_sDBUSState.failText.empty() ? std::nullopt : std::optional(m_sDBUSState.failText);
+}
+
+void CFingerprint::terminate() {
+    m_sDBUSState.connection->leaveEventLoop();
+}
```
PaideiaDilemma, line 55: need a check here?

PaideiaDilemma: `m_sDBUSState.connection` is a nullptr when fingerprint is not enabled right?

Kochanac:
Hello, tested this feature, here is what I think:

- when trying to use the wrong finger 3 times, hyprlock exits and unlocks the session (logs ([hyprlock -v](https://github.com/user-attachments/files/17355829/hyprlock_logs.txt)) and my config [hyprlock.conf.txt](https://github.com/user-attachments/files/17355837/hyprlock.conf.txt), compiled from a2ef6a9907f6c26720150af730cc81fed7f82688)

- maybe this is my config problem, but anyway: I don't see any error messages when getting verify-no-match error

- from a UX standpoint I would suggest that as an option you could add some message when starting to verify using fprint, so a user could see that the system is working correctly

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include "../helpers/Log.hpp"
+#include "src/config/ConfigManager.hpp"
+
+#include <filesystem>
+#include <unistd.h>
+#include <pwd.h>
+#include <security/pam_appl.h>
+#if __has_include(<security/pam_misc.h>)
+#include <security/pam_misc.h>
+#endif
```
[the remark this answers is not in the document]
moggiesir: Removed

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: It's not straightforward AFAICT since input-field is a special category according to the config so I'm not sure how to get it aside from looping through the widget configs and finding the input-field, if it's there, but let me know if I'm missing something.

moggiesir:
> Hello, tested this feature, here is what I think:
> 
> * when trying to use the wrong finger 3 times, hyprlock exits and unlocks the session (logs ([hyprlock -v](https://github.com/user-attachments/files/17355829/hyprlock_logs.txt)) and my config [hyprlock.conf.txt](https://github.com/user-attachments/files/17355837/hyprlock.conf.txt), compiled from [a2ef6a9](https://github.com/hyprwm/hyprlock/commit/a2ef6a9907f6c26720150af730cc81fed7f82688))

This is now fixed, it was a bad state during some refactoring.

> * probably this is my config problem, but anyway: I don't see any error messages in the input field when getting verify-no-match error

I didn't want to mess with the input-field so as to not introduce edge cases with the password input/error handling, so you can use $FPRINTMESSAGE in a label.

> * from a UX standpoint I would suggest that as an option you could add some message when starting to verify using fprint, so a user could see that the system is working correctly

I added an option general:fingerprint_ready_message that $FPRINTMESSAGE will display when ready for fingerprint input.

Kochanac:
Indeed the bug is fixed, thanks!

About fingerprint_ready_message, I meant another thing, sorry for the unclear explanation. I've meant that when you touch the scanner, the system should immediately display some message like "scanning..." to indicate that scan is started. I think it would be useful for improving responsiveness as sometimes the scan can be longer than 500ms

But it looks like fprint does not expose such interface, so it is probably impractical to do. fingerprint_ready_message is also useful, so I would leave it as is

src/core/Fingerprint.cpp
PaideiaDilemma: You are right.

Idea: Add `$FPRINTPROMPT` and `$FPRINTFAIL` like `$PROMPT` and `$FAIL` in `Auth.cpp` (`getLastPrompt` and `getLastFailText`), then we would not need the timer at all. Although you would not be able to have `$FPRINTFEEDBACK` like behavior. Currently you also don't have that for normal auth. Switching between fail and prompt is input-field exclusive.

PaideiaDilemma:
> I didn't want to mess with the input-field so as to not introduce edge cases with the password input/error handling, so you can use $FPRINTMESSAGE in a label.

In my opinion it would make sens to be able to have the fp feedback in the input-field, if it is empty. But for that we might want to add an abstraction so that we don't have to add logic for an additional auth module in the input-field. That could be done later.

moggiesir:
> About fingerprint_ready_message, I meant another thing, sorry for the unclear explanation. I've meant that when you touch the scanner, the system should immediately display some message like "scanning..." to indicate that scan is started. I think it would be useful for improving responsiveness as sometimes the scan can be longer than 500ms
> 
> But it looks like fprint does not expose such interface, so it is probably impractical to do. fingerprint_ready_message is also useful, so I would leave it as is

There's a property on the device for whether the finger is present, so I changed to use the PropertiesChanged signal to listen for finger presence and display a message at that time (general:fingerprint_present_message). On my device, it's pretty snappy to read my fingerprint, so I don't like the extra messages.

src/core/Fingerprint.cpp
moggiesir: There's a signal for finger presence, so I switched to change the $FPRINTMESSAGE at that time, removing the need for the timer.

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
+#include "Fingerprint.hpp"
+#include "../helpers/Log.hpp"
+#include "src/config/ConfigManager.hpp"
```
PaideiaDilemma, line 3: Relative include pls

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+CFingerprint::CFingerprint() {
+    static auto* const PFINGERPRINTREADY   = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:fingerprint_ready_message"));
+    m_sFingerprintReady                    = *PFINGERPRINTREADY;
+    static auto* const PFINGERPRINTPRESENT = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:fingerprint_present_message"));
+    m_sFingerprintPresent                  = *PFINGERPRINTPRESENT;
+}
+
+static void fingerprintUnlockCallback(std::shared_ptr<CTimer> self, void* data) {
+    g_pHyprlock->unlock();
```
PaideiaDilemma, line 26: I don't really like that, because i think the check for `m_bAuthenticated` should be right here. Just make it a `onFingerprintCheckTimer` like with the password authentication and call it whether or not we are authenticated.

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+    m_sDBUSState.connection = sdbus::createSystemBusConnection();
+
+    registerSleepHandler();
+
+    if (m_sDBUSState.sleeping) {
+        m_sDBUSState.connection->enterEventLoop();
+    } else if (startVerify()) {
+        m_sDBUSState.message = m_sFingerprintReady;
+        g_pHyprlock->enqueueForceUpdateTimers();
+        m_sDBUSState.connection->enterEventLoop();
```
PaideiaDilemma, line 66: NITPICK: would be a bit easier to understand if `enterEventLoop` would not happen in `setupDBUS`, but either in the std::thread ctor or in a dedicated `enterLoop` function.

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+    m_sDBUSState.login->uponSignal("PrepareForSleep").onInterface(LOGIN_MANAGER).call([this](bool start) {
+        Debug::log(LOG, "prepare for sleep");
+        if (start) {
+            m_sDBUSState.sleeping = true;
+            stopVerify();
+            m_sDBUSState.inhibitLock.reset();
+        } else {
+            m_sDBUSState.sleeping = false;
+            inhibitSleep();
+            std::this_thread::sleep_for(std::chrono::seconds(1));
```
PaideiaDilemma, line 89: Can you add a comment for why this sleep is needed?

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+    if (!m_sDBUSState.sleeping)
+        inhibitSleep();
+}
+
+void CFingerprint::inhibitSleep() {
+    m_sDBUSState.login->callMethod("Inhibit")
+        .onInterface(LOGIN_MANAGER)
+        .withArguments("sleep", "hyprlock", "Release fingerprint before sleep", "delay")
+        .storeResultsTo(m_sDBUSState.inhibitLock);
+}
```
PaideiaDilemma, line 106: So this is used to make sure we call `stopVerify` before we are going to sleep right?

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+        });
+
+    m_sDBUSState.device->finishRegistration();
+    return true;
+}
+
+void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
+    Debug::log(LOG, "Status: {} {}", done, result);
+    auto matchResult = s_mapStringToTestType[result];
+    if (m_sDBUSState.sleeping && matchResult != MatchResult::Disconnected) {
```
PaideiaDilemma, line 147: Brackets can be omitted.

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+            break;
+        case MatchResult::UnknownError:
+            stopVerify();
+            m_sDBUSState.message = "Unknown fingerprint error.";
+            break;
+        case MatchResult::Match:
+            stopVerify();
+            m_bAuthenticated     = true;
+            m_sDBUSState.message = "";
+            Debug::log(LOG, "matched");
```
PaideiaDilemma, line 171: This log and others could be improved. Ideally prefixed with `fprint: ` or smth.

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,240 @@
...
+        return;
+    }
+    switch (matchResult) {
+        case MatchResult::NoMatch:
+            stopVerify();
+            if (m_sDBUSState.retries < 2) {
+                startVerify();
+                done = false;
+                m_sDBUSState.retries++;
+                m_sDBUSState.message = RETRY_MESSAGE;
```
PaideiaDilemma, line 157: Check whether `startVerify` succeeded? otherwise that would cause a silent fail I think.

PaideiaDilemma:
The rest looks pretty good.
Thanks for working on this! Sadly I cannot properly test it, because my fp reader has no fprintd driver ;(

noxxious:
The PR needs changes to flake.nix to add extra buildInputs: sdbus-cpp and systemdLibs.

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: Cleaned this up a bit and moved the logic into the thread ctor.

src/core/Fingerprint.cpp
moggiesir: Right, updated the inhibit reason.

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: Done

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,232 @@
...
+        g_pHyprlock->unlock();
+}
+
+void CFingerprint::start() {
+    if (!m_bEnabled)
+        return;
+    std::thread([this]() {
+        // For grace or SIGUSR1 unlocks
+        if (g_pHyprlock->isUnlocked())
+            return;
```
PaideiaDilemma, line 38: That does not hurt but I think it is also not needed.

PaideiaDilemma (approved):

matejdro:
I have did a quick check of this using fingerprint sensor in the ThinkPad T14s (AMD) Gen 3 and everything seem to be working well. Thanks

src/config/ConfigManager.cpp
```diff
@@ -50,6 +50,9 @@ void CConfigManager::init() {
     m_config.addConfigValue("general:immediate_render", Hyprlang::INT{0});
     m_config.addConfigValue("general:pam_module", Hyprlang::STRING{"hyprlock"});
     m_config.addConfigValue("general:fractional_scaling", Hyprlang::INT{2});
+    m_config.addConfigValue("general:enable_fingerprint", Hyprlang::INT{0});
+    m_config.addConfigValue("general:fingerprint_ready_message", Hyprlang::STRING{""});
+    m_config.addConfigValue("general:fingerprint_present_message", Hyprlang::STRING{""});
```
PaideiaDilemma, line 55: add sensible defaults for the messages?

moggiesir: Done

src/core/Fingerprint.cpp
moggiesir: Removed

b5l: > That does not hurt
> **I think** it is also not needed

Unless you **know**, I consider this a reason for keeping those lines in.

PaideiaDilemma: well i know that **should not** be needed, because a race, where we go through that check and then hyprlock is unlocked, (via grace or USR1) might cause some problems. So even if the check fixes smth, it does not fix it properly.

PaideiaDilemma: @moggiesir is it fine when the main thread calls `terminate` before `m_sDBUSState.connection = sdbus::createSystemBusConnection();` succeeds? Also is `leaveEventLoop` thread safe?

moggiesir: It's fine when the main thread calls `terminate` before `m_sDBUSState.connection = sdbus::createSystemBusConnection();` succeeds since `terminate` checks if the pointer is initialized. I also made sure it's fine to call `leaveEventLoop` even if `enterEventLoop` was never called.

For `leaveEventLoop` thread safety, from the [docs](https://github.com/Kistler-Group/sdbus-cpp/blob/master/docs/using-sdbus-c++.md):

> An event loop that blocks in the synchronous enterEventLoop() call can be unblocked through leaveEventLoop() call on the respective bus connection issued from a different thread or from an OS signal handler.

Further, I changed to return the thread from `CFingerprint::start()` so that hyprlock will wait for the thread to terminate to make sure we don't try to destroy the dbus connection while some call is being made.

PaideiaDilemma: Nice, thanks!

PaideiaDilemma:
@vaxerski I think this looks good

src/core/Fingerprint.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+static void fingerprintCheckCallback(std::shared_ptr<CTimer> self, void* data) {
+    if (g_pFingerprint->isAuthenticated())
+        g_pHyprlock->unlock();
+}
+
+std::optional<std::thread> CFingerprint::start() {
+    if (!m_bEnabled)
+        return {};
+    return std::thread([this]() {
```
vaxerski, line 35: absolutely do _not_ multithread here. Get the pollfd from sdbus-cpp and integrate it within the main event loop (see xdph: https://github.com/hyprwm/xdg-desktop-portal-hyprland/blob/fb9c8d665af0588bb087f97d0f673ddf0d501787/src/core/PortalManager.cpp#L270)

CMakeLists.txt
```diff
@@ -56,7 +56,8 @@ pkg_check_modules(
   pangocairo
   libdrm
   gbm
-  hyprutils>=0.2.3)
+  hyprutils>=0.2.3
+  sdbus-c++)
```
vaxerski, line 60: you are using sdbus-c++ 2 here, so please add `>=2.0.0` to not confuse packagers (xdph still uses 1.x)

moggiesir: What am I using from sdbus-c++>=2.0.0? On my system, I only have 1.5.0 installed.

src/core/Fingerprint.cpp
moggiesir: Done. I was originally trying to maintain more consistency with CAuth.

It makes things a bit more complicated, but I kept it so that the dbus connection is only initiated/listened to if fingerprint auth is enabled. I'd be happy to change it to always create the connection if you'd rather that.

CMakeLists.txt
vaxerski: oh. I thought the syntax looked 2.0-like. Use 2.0 then. If you have a system 1.5, either install 2.0 or add a subproject and check it out to the latest release.

2.0 is breaking and I don't want to "fix it later", already lazy to do it with xdph

moggiesir: Done.

src/core/Fingerprint.hpp
```diff
@@ -0,0 +1,70 @@
...
+  public:
+    CFingerprint();
+
+    std::shared_ptr<sdbus::IConnection> start();
+    bool                                isAuthenticated();
+    std::optional<std::string>          getLastMessage();
+    void                                terminate();
+
+  private:
+    enum class MatchResult {
```
vaxerski, line 21: C++: `enum`
Also, please:
- call enum variables in CAPS and with an identifiable prefix (e.g. `MATCH_NO_MATCH`, `MATCH_RETRY`)
- add a variable for invalid e.g. `MATCH_INVALID = 0`
- have the _first_ enum value be defined (e.g. as above = 0)

src/core/Fingerprint.hpp
```diff
@@ -0,0 +1,70 @@
...
+        int                                 retries  = 0;
+        bool                                sleeping = false;
+    } m_sDBUSState;
+
+    std::string                        m_sFingerprintReady;
+    std::string                        m_sFingerprintPresent;
+    bool                               m_bAuthenticated = false;
+    bool                               m_bEnabled       = false;
+
+    std::map<std::string, MatchResult> s_mapStringToTestType = {{"verify-no-match", MatchResult::NoMatch},
```
vaxerski, line 50: can this be moved to .cpp and static'd?

src/core/hyprlock.cpp
```diff
@@ -425,16 +430,21 @@ void CHyprlock::run() {
...
-        {
-            .fd     = wl_display_get_fd(m_sWaylandState.display),
-            .events = POLLIN,
-        },
+    pollfd pollfds[2];
+    pollfds[0] = {
+        .fd     = wl_display_get_fd(m_sWaylandState.display),
+        .events = POLLIN,
     };
+    if (conn)
```
vaxerski, line 438: {}, this is not a short if

vaxerski (commented):
minor nitpicks, rest is alr, needs a wiki mr

src/core/hyprlock.cpp
moggiesir: Done

src/core/Fingerprint.hpp
moggiesir: Done

src/core/Fingerprint.hpp
moggiesir: Done

moggiesir:
Opened https://github.com/hyprwm/hyprland-wiki/pull/816 for the wiki

moggiesir:
Yes, with the new parallel auth support, users should go back to the default pam file.

src/core/Fingerprint.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "hyprlock.hpp"
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <condition_variable>
```
vaxerski, line 8: include not needed anymore?

vaxerski:
apart from that lgtm

src/core/Fingerprint.hpp
moggiesir: Done

PaideiaDilemma:
@vaxerski

Quite a few fixes since last release. I think a hyprlock release is due.
Wanna do that before or after this?

vaxerski:
I need @fufexan to fix nix ci and then we can release

vaxerski:
man I love fufexan

PR #515, opened 2024-10-12
misc: use Vector2D, Box and Mat3x3 from hyprutils

PaideiaDilemma:
Switch to hyprutils for Vector2D, Box and Mat3x3.


src/core/LockSurface.cpp
```diff
@@ -7,7 +7,7 @@
 
 static void handleConfigure(void* data, ext_session_lock_surface_v1* surf, uint32_t serial, uint32_t width, uint32_t height) {
     const auto PSURF = (CSessionLockSurface*)data;
-    PSURF->configure({width, height}, serial);
+    PSURF->configure({(double)width, (double)height}, serial);
```
PaideiaDilemma, line 10: This ok?

src/helpers/Math.hpp
```diff
@@ -0,0 +1,12 @@
...
+#include <wayland-client.h>
+
+#include <hyprutils/math/Box.hpp>
+#include <hyprutils/math/Vector2D.hpp>
+#include <hyprutils/math/Mat3x3.hpp>
+
+using namespace Hyprutils::Math;
+
+eTransform          wlTransformToHyprutils(wl_output_transform t);
+wl_output_transform invertTransform(wl_output_transform tr);
```
PaideiaDilemma, line 12: Same as in hyprland. I think we don't want to move `wlTransformToHyprutils` and `invertTransform` to hyprutils, as that would mean we have to get the `WL_OUTPUT_TRANSFORM*` definitions from somewhere right?

Also `Math.hpp` is included for Box, Vector2D and Mat3x3. But clangd complains that the header is not used directly. Let me know if we want the includes to work differently.

src/renderer/widgets/IWidget.hpp
```diff
@@ -1,8 +1,10 @@
 #pragma once
 
-#include "../../helpers/Vector2D.hpp"
+#include "../../helpers/Math.hpp"
 #include <string>
 
+#define VEC2CONVERT(vec) Vector2D(vec.x, vec.y)
+
```
PaideiaDilemma, line 7: This is just temporary.
What would be the best way to convert Hyprlang::VEC2 to Vector2D in all the widgets?

src/renderer/widgets/IWidget.cpp
```diff
@@ -20,12 +20,18 @@ namespace std {
 }
 #endif
 
+Vector2D rotateVector(const Vector2D& vec, const double& ang) {
+    const double COS = std::abs(std::cos(ang));
+    const double SIN = std::abs(std::sin(ang));
+    return Vector2D(vec.x * COS + vec.y * SIN, vec.x * SIN + vec.y * COS);
+}
+
```
PaideiaDilemma, line 28: We could move that to hyprutils I guess.

src/core/LockSurface.cpp
vaxerski: zis ok

src/helpers/Math.hpp
vaxerski: looks fine

src/renderer/widgets/IWidget.hpp
vaxerski: write a custom ctor out-of-line? is that allowed in C++? if not, that or a fn.

src/renderer/widgets/IWidget.cpp
vaxerski: yes, but for now it has to stay here

src/renderer/widgets/IWidget.hpp
PaideiaDilemma: you can't sadly. I added a fn.

vaxerski (approved):
lgtm

vaxerski:
anything else?

PaideiaDilemma:
hmm nah i think this is it.

PaideiaDilemma:
ahh wait maybe add `hyprutils>=0.2.3` to the readme

vaxerski:
nah who cares about the readme

PaideiaDilemma:
not me xD

vaxerski:
I've doen it myslfe

Issue #519, opened 2024-10-18
Crash when using $TIME


2Belette:
### Regression?

Yes

### Hyprlock Info and Version

hyprlock-0.4.1 @ FreeBSD

<details> <summary>Hyprlock config</summary>

```sh
label {
    monitor =
#    text = cmd[update:1000] echo "$(date +"%H:%M:%S")"
    text = $TIME
    color = rgba(200, 200, 200, 1.0)
    font_size = 60
    font_family = Noto Sans
    position = -30, 0
    halign = right
    valign = top
}
```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch main at commit 4520b30d498daca8079365bdb909a8dea38e8d55 portbld (?).
Date: Wed Oct  9 12:54:39 UTC 2024
Tag: v0.44.1, commits: ?
built against aquamarine 0.4.3


flags: (if any)


System Information:
System name: FreeBSD
Node name: xps
Release: 14.1-RELEASE-p5
Version: FreeBSD 14.1-RELEASE-p5 GENERIC


GPU information: 
vgapci0@pci0:0:2:0:	class=0x030000 rev=0x0c hdr=0x00 vendor=0x8086 device=0x46a6 subvendor=0x1028 subdevice=0x0af3
    vendor     = 'Intel Corporation'
    device     = 'Alder Lake-P GT2 [Iris Xe Graphics]'
    class      = display
    subclass   = VGA


os-release: NAME=FreeBSD
VERSION="14.1-RELEASE-p5"
VERSION_ID="14.1"
ID=freebsd
ANSI_COLOR="0;31"
PRETTY_NAME="FreeBSD 14.1-RELEASE-p5"
CPE_NAME="cpe:/o:freebsd:freebsd:14.1"
HOME_URL="https://FreeBSD.org/"
BUG_REPORT_URL="https://bugs.FreeBSD.org/"


plugins:
```

</details>

### Description

This is crashing hyprlock with no hint into logging nor verbose mode.
If I comment the line with $TIME and uncommenting the line above, everything is working well.
It was working before.

### How to reproduce

using $TIME

PaideiaDilemma:
You probably did not set up your timezone.
I opened #527 to handle this a bit more gracefully.

2Belette:
Thanks !
[...]

PR #527, opened 2024-10-24
widgets: check current_zone pointer

PaideiaDilemma:
Just to be sure we don't deref that if it is not set.

Probably fixes https://github.com/hyprwm/hyprlock/issues/519

### DIFF
--- a/src/helpers/Log.hpp
+++ b/src/helpers/Log.hpp
@@ -18,7 +18,6 @@ enum eLogLevel {
         Debug::log(CRIT, "\n==========================================================================================\nASSERTION FAILED! \n\n{}\n\nat: line {} in {}",            \
                    std::format(reason, ##__VA_ARGS__), __LINE__,                                                                                                                   \
                    ([]() constexpr -> std::string { return std::string(__FILE__).substr(std::string(__FILE__).find_last_of('/') + 1); })().c_str());                               \
-        printf("Assertion failed! See the log in /tmp/hypr/hyprland.log for more info.");                                                                                          \
         std::abort();                                                                                                                                                              \
     }
 
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -113,9 +113,8 @@ static std::string getTime() {
             logMissingTzOnce = false;
         }
         hhmmss = std::chrono::hh_mm_ss{TPNOW - std::chrono::floor<std::chrono::days>(TPNOW)};
-    } else {
+    } else
         hhmmss = std::chrono::hh_mm_ss{PCURRENTTZ->to_local(TPNOW) - std::chrono::floor<std::chrono::days>(PCURRENTTZ->to_local(TPNOW))};
-    }
 
     const auto HRS  = hhmmss.hours().count();
     const auto MINS = hhmmss.minutes().count();
@@ -134,9 +133,8 @@ static std::string getTime12h() {
             logMissingTzOnce = false;
         }
         hhmmss = std::chrono::hh_mm_ss{TPNOW - std::chrono::floor<std::chrono::days>(TPNOW)};
-    } else {
+    } else
         hhmmss = std::chrono::hh_mm_ss{PCURRENTTZ->to_local(TPNOW) - std::chrono::floor<std::chrono::days>(PCURRENTTZ->to_local(TPNOW))};
-    }
 
     const auto HRS  = hhmmss.hours().count();
     const auto MINS = hhmmss.minutes().count();
--- a/src/renderer/widgets/IWidget.cpp
+++ b/src/renderer/widgets/IWidget.cpp
@@ -100,21 +100,46 @@ static void replaceAllLayout(std::string& str) {
     }
 }
 
+static bool        logMissingTzOnce = true;
 static std::string getTime() {
-    const auto current_zone = std::chrono::current_zone();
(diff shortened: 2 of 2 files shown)


src/renderer/widgets/IWidget.cpp
```diff
@@ -100,21 +100,26 @@ static void replaceAllLayout(std::string& str) {
...
 
+static const auto  PCURRENTTZ = std::chrono::current_zone();
 static std::string getTime() {
-    const auto current_zone = std::chrono::current_zone();
-    const auto HHMMSS       = std::chrono::hh_mm_ss{current_zone->to_local(std::chrono::system_clock::now()) -
-                                              std::chrono::floor<std::chrono::days>(current_zone->to_local(std::chrono::system_clock::now()))};
-    const auto HRS          = HHMMSS.hours().count();
-    const auto MINS         = HHMMSS.minutes().count();
+    if (!PCURRENTTZ)
+        return "TIMEZONE UNKNOWN";
```
vaxerski, line 106: JUST RETURN UTC

vaxerski: MAYBE LOG IT ONCE

PaideiaDilemma: > JUST RETURN UTC

Reasonable, done.

> MAYBE LOG IT ONCE

Turns out the current_zone pointer is not guarantied to be static i from looking at the sources. So it is much easier to always log it. I think it's fine to spam this once in ~30 secs. current_zone being a nullptr shouldn't happen really.

src/helpers/Log.hpp
```diff
@@ -18,7 +18,6 @@ enum eLogLevel {
         Debug::log(CRIT, "\n==========================================================================================\nASSERTION FAILED! \n\n{}\n\nat: line {} in {}",            \
                    std::format(reason, ##__VA_ARGS__), __LINE__,                                                                                                                   \
                    ([]() constexpr -> std::string { return std::string(__FILE__).substr(std::string(__FILE__).find_last_of('/') + 1); })().c_str());                               \
-        printf("Assertion failed! See the log in /tmp/hypr/hyprland.log for more info.");                                                                                          \
```
vaxerski, line 21: ??

src/renderer/widgets/IWidget.cpp
```diff
@@ -101,20 +101,34 @@ static void replaceAllLayout(std::string& str) {
...
-    const auto current_zone = std::chrono::current_zone();
-    const auto HHMMSS       = std::chrono::hh_mm_ss{current_zone->to_local(std::chrono::system_clock::now()) -
-                                              std::chrono::floor<std::chrono::days>(current_zone->to_local(std::chrono::system_clock::now()))};
-    const auto HRS          = HHMMSS.hours().count();
-    const auto MINS         = HHMMSS.minutes().count();
+    const auto PCURRENTTZ = std::chrono::current_zone();
+    const auto PUTC       = std::chrono::locate_zone("UTC");
+
+    if (!PCURRENTTZ) {
+        Debug::log(WARN, "Current timezone unknown for $TIME. Using UTC!");
```
vaxerski, line 108: this will be spammed, only log it once

```cpp
static bool once = true;
```

src/helpers/Log.hpp
PaideiaDilemma: Since Debug::log will log to stdout and the info about `/tmp/hypr/hyprland.log` is not accurate this printf seemed useless.
Removed it because i added an RASSERT for PUTC.

src/renderer/widgets/IWidget.cpp
```diff
@@ -100,21 +100,40 @@ static void replaceAllLayout(std::string& str) {
...
-    const auto MINS         = HHMMSS.minutes().count();
+    const auto PCURRENTTZ = (true) ? nullptr : std::chrono::current_zone();
+    const auto PUTC       = std::chrono::locate_zone("UTC");
+
+    if (!PCURRENTTZ && logMissingTzOnce) {
+        Debug::log(WARN, "Current timezone unknown for $TIME. Falling back to UTC!");
+        logMissingTzOnce = false;
+    }
+
+    RASSERT(PCURRENTTZ || PUTC, "Failed to get fallback UTC timezone");
```
PaideiaDilemma, line 113: Crash or just return a generic string?

vaxerski: can't you use system clock? isnt that guaranteed?

PaideiaDilemma:
If you want i can move the commit about the RASSERT printf to another PR. Just let me know : )

src/renderer/widgets/IWidget.cpp
```diff
@@ -100,21 +100,46 @@ static void replaceAllLayout(std::string& str) {
...
+
+    //
+    std::chrono::hh_mm_ss<std::chrono::system_clock::duration> hhmmss;
+    if (!PCURRENTTZ) {
+        if (logMissingTzOnce) {
+            Debug::log(WARN, "Current timezone unknown for $TIME. Falling back to UTC!");
+            logMissingTzOnce = false;
+        }
+        hhmmss = std::chrono::hh_mm_ss{TPNOW - std::chrono::floor<std::chrono::days>(TPNOW)};
+    } else {
```
vaxerski, line 116: lmao man put {} around a short if

vaxerski (commented):
rest ok

src/renderer/widgets/IWidget.cpp
PaideiaDilemma: else gets me sometimes...

vaxerski (approved):

Issue #535, opened 2024-10-28
Howdy's pam module causes unlocks when using reload commands (security risk! DONT use Howdy for Lockscreen unlocks!)


londeril:
### Regression?

No

### Hyprlock Info and Version

Hyprlock version 0.5.0
<details> <summary>Hyprlock config</summary>

```sh
# Hyprlock                                                          
# 

general {
}

background {
    monitor =
    path = screenshot  
    blur_size = 5
    blur_passes = 1 
    noise = 0.0117
    contrast = 1.3000
    brightness = 0.8000
    vibrancy = 0.2100
    vibrancy_darkness = 0.0
}

input-field {
    monitor =
    size = 250, 50
    outline_thickness = 3
    dots_size = 0.33 
    dots_spacing = 0.15
    dots_center = true
    outer_color = rgba(B9A100ff)  
    inner_color = $color0
    font_color = $color12
    fade_on_empty = true
    placeholder_text = <i>Magic Word...</i>
    hide_input = false
    position = 0, 200
    halign = center
    valign = bottom
}

image {
    monitor =
    path = ~/.dotfiles/lockscreen.jpg
    size = 1000
    rounding = 10
    border_size = 2
    border_color = rgba(B9A100ff)
    rotate = 0
    reload_time = 10
    reload_cmd = find ~/Pictures/LandscapeSaver/ -type f | shuf -n 1
    position = 0, 0
    halign = center
    valign = center
}
```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 4520b30d498daca8079365bdb909a8dea38e8d55  (version: bump to 0.44.1).
Date: Wed Oct 9 12:54:39 2024
Tag: v0.44.1, commits: 5309
built against aquamarine 0.4.3


flags: (if any)


System Information:
System name: Linux
Node name: nova
Release: 6.11.5-zen1-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Tue, 22 Oct 2024 18:31:33 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] [8086:9a49] (rev 01) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
[...]

PR #536, opened 2024-10-28
core: lock m_vTimers for forceUpdateTimers

PaideiaDilemma:
Previously `m_vTimers` was not locked when calling `handleForceUpdateSinal`. Don't know how but that might be the culprit behind #535.
To avoid problems with async signals, I introduced this atomtic variable and integrated `forceUpdateTimers` into the main loop instead of locking `m_vTimers` via a signal.


PaideiaDilemma:
Need to revisit i think this would cause problems when we get no poll events

PR #539, opened 2024-10-30
core: improve signal handling

### LINKED ISSUE #535: Howdy's pam module causes unlocks when using reload commands (security risk! DONT use Howdy for Lockscreen unlocks!)

PaideiaDilemma:
This PR tries to handle SIGUSR1, SIGUSR2 and termination signals via the main thread using `signalfd`.

I did that cause when we `forceUpdateTimers`, we are currently not safe in multiple ways.
We don't acquire the timers lock for `m_vTimers`, which we can't do cause we are in a async signal context.
We call other async signal unsafe functions.
So I tried moving `forceUpdateTimers` to the main thread (#536), but that has the problem that we need to wake the main thread somehow and that would require async signal unsafe functions again.

Considering hyprlock is event driven using signalfd made sense.

I hope that this fixes #535.

Some testing wanted before merging. 
Thanks!


src/core/hyprlock.cpp
```diff
@@ -1156,7 +1226,7 @@ zwlr_screencopy_manager_v1* CHyprlock::getScreencopy() {
 }
 
 void CHyprlock::attemptRestoreOnDeath() {
-    if (m_bTerminate || m_sCurrentDesktop != "Hyprland")
```
PaideiaDilemma, line 1159: Any Idea why we didn't try to restore when `m_bTerminate` is true before?

PaideiaDilemma: Maybe we want to check if we are actually locked instead?
Cause if we aren't there is no point in trying to restore.

PaideiaDilemma:
Needs a bit more work

src/core/hyprlock.cpp
vaxerski: uhhh cuz if it crashes after nlock

PaideiaDilemma: yeah but terminate does not mean we already unlocked

vaxerski: oh then ok

PR #554, opened 2024-11-12
widgets: render frambuffers with alpha 1.0

PaideiaDilemma:
I broke shape and image borders on accident by using `data.opacity`.
Did not realize it because i don't use fade-in.

CC @alba4k


PaideiaDilemma:
CColor commit is just cause it fixes debug prints of Gradient colors.

PR #573, opened 2024-12-02
Add password hash authentication

spacefrogg:
As an alternative to PAM authentication, password hash authentication only relies on the availability of libgcrypt (and some program like OpenSSL or sha256sum to create the hash).

Supports salted hashes and all hash algorithms that are available in the actual libgcrypt installation.

This is practical in more restricted environments. As an example: I work on a Ubuntu machine w/o root privileges and use Hyprland via Nix Home Manager. Ubuntu uses a patched PAM library to support their own `@include` primitive which is incompatible with standard PAM. Additionally `dlopen` across PAM library version does not work well from 1.6 to 1.5.

I am aware that the documentation likely has to move from the README to the Wiki, but I put it there for easy initial review.

### DIFF
--- a/src/core/PwAuth.cpp
+++ b/src/core/PwAuth.cpp
@@ -1,0 +1,191 @@
+#include "PwAuth.hpp"
+#include "hyprlock.hpp"
+#include "../helpers/Log.hpp"
+#include "../config/ConfigManager.hpp"
+#include <cstddef>
+#include <cstring>
+#define GCRYPT_NO_DEPRECATED
+#define GCRYPT_NO_MPI_MACROS
+#define NEED_LIBGCRYPT_VERSION nullptr
+#include <gcrypt.h>
+
+using namespace std::chrono_literals;
+
+static std::unique_ptr<unsigned char[]> hex2Bytes(const std::string& hex) noexcept {
+    auto bytes = std::make_unique<unsigned char[]>(hex.length() / 2);
+    for (std::size_t i = 0; i < hex.length() / 2; ++i) {
+        try {
+            auto v = std::stoi(hex.substr(2 * i, 2), nullptr, 16);
+            if (v >= 0)
+                bytes[i] = static_cast<unsigned char>(v);
+            else
+                throw std::invalid_argument("invalid hex value");
+        } catch (std::invalid_argument const& e) {
+            Debug::log(ERR, "auth: invalid password_hash");
+            bytes = nullptr;
+        } catch (std::out_of_range const& e) {
+            // Should never happen, as 2-byte substrings should never go o-o-r.
+            Debug::log(CRIT, "auth: implementation error in hex2Bytes conversion");
+            bytes = nullptr;
+        }
+    }
+    return bytes;
+}
+
+static std::string bytes2Hex(const unsigned char* bytes, std::size_t len) {
+    std::stringstream ss;
+    ss << std::setw(2) << std::setfill('0') << std::hex;
+    for (std::size_t i = 0; i < len; ++i)
+         ss << (int)bytes[i];
+    return ss.str();
+}
+
+CPwAuth::CPwAuth() {
+
+    if (gcry_check_version(NEED_LIBGCRYPT_VERSION))
+        gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
+    else
+        Debug::log(CRIT, "libgcrypt too old");
+
+    if (gcry_control(GCRYCTL_INITIALIZATION_FINISHED_P)) {
+
+        // Handle the hash algorithm
+        static auto const ALGO = *(Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:hash_algorithm"));
+        m_iAlgo                 = gcry_md_map_name(ALGO);
+        m_iDigestLen            = gcry_md_get_algo_dlen(m_iAlgo);
+        if (m_iAlgo) {
+            static auto const err = gcry_err_code(gcry_md_test_algo(m_iAlgo));
+            if (err == GPG_ERR_NO_ERROR) {
+
+                // Handle the salt
+                static auto* const SALT = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:hash_salt"));
+                m_szSalt                = std::string(*SALT);
+
+                // Handle the expected hash
+                static auto* const HASH = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:password_hash"));
+                static auto const  hash = std::string(*HASH);
+                if (hash.empty() || (hash.size() % 2) || (hash.length() != 2uL * m_iDigestLen)) {
+                    Debug::log(ERR, "auth: password_hash has incorrect length for algorithm {} (got: {}, expected: {})", ALGO, hash.size(), 2uL * m_iDigestLen);
+                    m_bLibFailed = true;
+                } else {
+                    m_aHash = hex2Bytes(hash);
+                    if (!m_aHash || hash.empty())
+                        m_bLibFailed = true;
+                }
+            } else {
+                // Might be due to FIPS mode
+                Debug::log(CRIT, "auth: hash algorithm unavailable: {}", ALGO);
+                m_bLibFailed = true;
+            }
+        } else {
+            Debug::log(ERR, "auth: unknown hash algorithm: {}", ALGO);
+            m_bLibFailed = true;
+        }
+    } else {
+        Debug::log(CRIT, "libgcrypt could not be initialized");
+        m_bLibFailed = true;
+    }
+}
+
+static void passwordCheckTimerCallback(std::shared_ptr<CTimer> self, void* data) {
+    g_pHyprlock->onPasswordCheckTimer();
+}
+
+void CPwAuth::start() {
+    std::thread([this]() {
+        reset();
+
+        waitForInput();
+
+        // For grace or SIGUSR1 unlocks
+        if (g_pHyprlock->isUnlocked())
+            return;
+
+        const auto AUTHENTICATED = auth();
+        m_bAuthenticated         = AUTHENTICATED;
+
+        if (g_pHyprlock->isUnlocked())
+            return;
+
+        g_pHyprlock->addTimer(1ms, passwordCheckTimerCallback, nullptr);
+    }).detach();
+}
+
+bool CPwAuth::auth() {
+    if (m_bLibFailed)
+        return true;
+
+    bool verdict;
+    auto digest  = std::make_unique<unsigned char[]>(m_iDigestLen);
+    auto istr    = m_sState.input;
+    istr.append(m_szSalt);
+
+    gcry_md_hash_buffer(m_iAlgo, digest.get(), istr.c_str(), istr.size());
+    Debug::log(TRACE, "auth: resulting hash {}", bytes2Hex(digest.get(), m_iDigestLen));
+    Debug::log(TRACE, "auth:  expected hash {}", bytes2Hex(m_aHash.get(), m_iDigestLen));
+    verdict = !std::memcmp(m_aHash.get(), digest.get(), m_iDigestLen);
+
+    if (verdict)
+        Debug::log(LOG, "auth: authenticated");
+    else
+        Debug::log(ERR, "auth: unsuccessful");
+
+    m_sState.authenticating = false;
+    /// DEBUG Code; replace constant with verdict
+    return verdict;
+}
+
+bool CPwAuth::isAuthenticated() {
+    return m_bAuthenticated;
+}
+
+// clearing the input must be done from the main thread
+static void clearInputTimerCallback(std::shared_ptr<CTimer> self, void* data) {
+    g_pHyprlock->clearPasswordBuffer();
+}
+
+void CPwAuth::waitForInput() {
+    g_pHyprlock->addTimer(1ms, clearInputTimerCallback, nullptr);
+    if (m_bLibFailed)
+        return;
+
+    std::unique_lock<std::mutex> lk(m_sState.inputMutex);
+    m_bBlockInput           = false;
+    m_sState.inputRequested = true;
+    m_sState.inputSubmittedCondition.wait(lk, [this] { return !m_sState.inputRequested || g_pHyprlock->m_bTerminate; });
+    m_bBlockInput = true;
+}
+
+void CPwAuth::submitInput(std::string input) {
+    std::unique_lock<std::mutex> lk(m_sState.inputMutex);
+    if (!m_sState.inputRequested)
+        Debug::log(ERR, "SubmitInput called, but the auth thread is not waiting for input!");
+    m_sState.input          = input;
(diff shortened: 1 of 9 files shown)


vaxerski (commented):
- what does libgcrypt offer over libcrypto?
- is there not a C++ library?

spacefrogg:
> * what does libgcrypt offer over libcrypto?
Hyprlock currently uses neither and I have just picked one of the two. Libgcrypt is much smaller and simpler than libcrypto as a whole. Although, mere one-off digest creation and checking looks quite similar in both. I knrew how to do it in libgcrypt. So I used it.

>     * is there not a C++ library?
I don't know of any famous and battle-proven library. Do you? So I stayed on the safe side. As the task is a purely functional "take this data, compute the digest and compare it to this other digest", there is no inherent value in a library with an object model. There is no state-transfer into libgcrypt which could collide with the C++ object model.

vaxerski:
I just am not a fan of `#define GCRYPT_NO_DEPRECATED` and C-style code in general, but fair, if there is no better alternative

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <cstddef>
+#include <cstring>
+#define GCRYPT_NO_DEPRECATED
+#define GCRYPT_NO_MPI_MACROS
+#define NEED_LIBGCRYPT_VERSION nullptr
+#include <gcrypt.h>
+
+using namespace std::chrono_literals;
+
+static auto hex2Bytes(const std::string& hex) -> std::unique_ptr<unsigned char[]> {
```
vaxerski, line 15: we don't do auto (...) -> type, please just static type fn(args)

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+    auto bytes = std::make_unique<unsigned char[]>(hex.length() / 2u);
```
vaxerski, line 16: what the actual fuck? std::vector??

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+        } catch (std::out_of_range const& e) {
+            // Should never happen, as 2-byte substrings should never go o-o-r.
+            Debug::log(CRIT, "auth: implementation error in hex2Bytes conversion");
+            bytes = nullptr;
+        }
+    }
+    return bytes;
+}
+
+static auto bytes2Hex(const unsigned char* bytes, std::size_t len) -> std::string {
```
vaxerski, line 36: same as before

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+    std::stringstream ss;
+    ss << std::setw(2) << std::setfill('0') << std::hex;
+    for (std::size_t i = 0; i < len; ++i)
+         ss << (int)bytes[i];
+    return ss.str();
```
vaxerski, line 41: std::format?

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+        gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
+    else
+        Debug::log(CRIT, "libgcrypt too old");
+
+    if (gcry_control(GCRYCTL_INITIALIZATION_FINISHED_P)) {
+
+        // Handle the hash algorithm
+        static auto* const ALGO = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:hash_algorithm"));
+        static auto const  algn = std::string(*ALGO);
+        m_iAlgo                 = gcry_md_map_name(algn.c_str());
```
vaxerski, line 56: I don't think you ever use the fact it's a std::string. Why not just use *ALGO here instead of c_str()

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+        m_iDigestLen            = gcry_md_get_algo_dlen(m_iAlgo);
+        if (m_iAlgo) {
+            static auto const err = gcry_err_code(gcry_md_test_algo(m_iAlgo));
+            if (err == GPG_ERR_NO_ERROR) {
+
+                // Handle the salt
+                static auto* const SALT = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:hash_salt"));
+                m_sSalt                 = std::string(*SALT);
```
vaxerski, line 64: explicit cast redundant

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+                // Handle the expected hash
+                static auto* const HASH = (Hyprlang::STRING*)(g_pConfigManager->getValuePtr("general:password_hash"));
+                static auto const  hash = std::string(*HASH);
+                if (hash.empty() || (hash.size() % 2) || (hash.length() != 2uL * m_iDigestLen)) {
```
vaxerski, line 69: uL redundant

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+        g_pHyprlock->addTimer(1ms, passwordCheckTimerCallback, nullptr);
+    }).detach();
+}
+
+bool CPwAuth::auth() {
+    if (m_bLibFailed)
+        return true;
+
+    bool verdict;
+    auto digest  = std::make_unique<unsigned char[]>(m_iDigestLen);
```
vaxerski, line 121: again, WTF?

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+        g_pHyprlock->addTimer(1ms, passwordCheckTimerCallback, nullptr);
+    }).detach();
+}
+
+bool CPwAuth::auth() {
+    if (m_bLibFailed)
+        return true;
+
+    bool verdict;
```
vaxerski, line 120: !!!!!!!!!!! uninitialized variable

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+    Debug::log(TRACE, "auth:  expected hash {}", bytes2Hex(m_sHash.get(), m_iDigestLen));
+    verdict = !std::memcmp(m_sHash.get(), digest.get(), m_iDigestLen);
+
+    if (verdict)
+        Debug::log(LOG, "auth: authenticated");
+    else
+        Debug::log(ERR, "auth: unsuccessful");
+
+    m_sState.authenticating = false;
+    /// DEBUG Code; replace constant with verdict
```
vaxerski, line 136: old comment?

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+        Debug::log(ERR, "SubmitInput called, but the auth thread is not waiting for input!");
+    m_sState.input          = input;
+    m_sState.inputRequested = false;
+    m_sState.authenticating = true;
+    m_sState.inputSubmittedCondition.notify_all();
+}
+
+std::optional<std::string> CPwAuth::getLastPrompt() {
+    std::string pmpt = "Password: ";
+    return std::optional(pmpt);
```
vaxerski, line 173: redundant explicit ctor

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+std::optional<std::string> CPwAuth::getLastFailText() {
+    std::string ret = "Password incorrect";
+    return std::optional(ret);
```
vaxerski, line 178: same

src/core/PwAuth.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  private:
+    SState                           m_sState;
+    bool                             m_bBlockInput    = true;
+    bool                             m_bAuthenticated = false;
+    bool                             m_bLibFailed     = false;
+    std::unique_ptr<unsigned char[]> m_sHash;
+    std::string                      m_sSalt;
+    int                              m_iAlgo;
+    unsigned int                     m_iDigestLen;
```
vaxerski, line 44: !!!! two uninitialized primitives!!!

src/core/PwAuth.hpp
```diff
@@ -0,0 +1,47 @@
...
+    bool                       checkWaiting();
+    void                       terminate();
+
+  private:
+    SState                           m_sState;
+    bool                             m_bBlockInput    = true;
+    bool                             m_bAuthenticated = false;
+    bool                             m_bLibFailed     = false;
+    std::unique_ptr<unsigned char[]> m_sHash;
+    std::string                      m_sSalt;
```
vaxerski, line 42: sz not s, blame the hungarians

src/core/PwAuth.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+    bool                       checkWaiting();
+    void                       terminate();
+
+  private:
+    SState                           m_sState;
+    bool                             m_bBlockInput    = true;
+    bool                             m_bAuthenticated = false;
+    bool                             m_bLibFailed     = false;
+    std::unique_ptr<unsigned char[]> m_sHash;
```
vaxerski, line 41: again wtf?

PaideiaDilemma:
I think pam is fundamentally broken, so I am not entirely opposed to a custom Auth method.

Two security relevant things.

- hashing is too fast. we need an artificial delay (could be configurable). Otherwise brute forcing trivial passwords is too easy.
-  I think we should store password hash + salt in a separate hyprlang config (like "secret.conf") and make sure it's permissions are -rw-r---- .

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+                Debug::log(CRIT, "auth: hash algorithm unavailable: {}", algn);
+                m_bLibFailed = true;
+            }
+        } else {
+            Debug::log(ERR, "auth: unknown hash algorithm: {}", algn);
+            m_bLibFailed = true;
+        }
+    } else {
+        Debug::log(CRIT, "libgcrypt could not be initialized");
+        m_bLibFailed = true;
```
PaideiaDilemma, line 88: Why not just crash? Maybe with a `RASSERT` or something. Without a crash you won't be able to log in, so I think that would make more sense.

vaxerski: +1

src/core/BAuth.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <memory>
+#include <optional>
+#include <string>
+
+class CBAuth {
```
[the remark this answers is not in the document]
spacefrogg: Sure, I was not particularly aware of your naming style and just mimicked what I saw. I will rename it to `IAuth`.

src/core/PwAuth.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "../helpers/Log.hpp"
+#include "../config/ConfigManager.hpp"
+#include <cstddef>
+#include <cstring>
+#define GCRYPT_NO_DEPRECATED
+#define GCRYPT_NO_MPI_MACROS
+#define NEED_LIBGCRYPT_VERSION nullptr
+#include <gcrypt.h>
+
+using namespace std::chrono_literals;
```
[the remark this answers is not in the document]
spacefrogg: Again, I have no strong feelings about this, but what is to be had against readable literals? Is it just a too un-C++ thing to do? I will gladly change it back, if you want me to.

src/core/PwAuth.cpp
spacefrogg: This happens during initialisation. You want the locker to lock up your system, because of an error that is outside of the user's realm of influence? Was thinking that the locker should rather not lock the system if it knows beforehand that it cannot reliably unlock afterwards.

In the end this is your decision to make. I just wanted to bring this up as a counter argument, as I have thought about this as well and came to the opposite conclusion.

spacefrogg:
What I just wrote is, of course, not equally true, for hashes that are used without salt.

src/core/PwAuth.cpp
spacefrogg: True, I was too lazy to find out what `Hyprlang::STRING*` was actually referring to.

src/core/PwAuth.cpp
spacefrogg: Without it, it's a `void*` which is not accepted by `std::string()`.

PaideiaDilemma:
> Is that not true for PAM as well? So, it should be some general artificial delay between login tries, not just for this method.

All PAM login configurations i came across have such a delay.

> Well, I don't believe in security by obscurity. The strength of the hash is usually as good as any public key. If you entrust public keys to the public, you can as well with the hash. I know, though, that other people think different about this and leave this for you to decide.

No!! Password hashes need to be treated with care. Especially if the salt is also stored in the same file. You are right that having the password hash leaked to another user does not straight up mean they can login. But with enough time you might be able to find the password (for example, if the password is in rockyou.txt and you know the salt, it is game over).

Finding a private key for a corresponding public key is MUCH harder than finding the plaintext for a certain hash. For that your security is only the plaintext length.

src/core/PwAuth.cpp
PaideiaDilemma: If we don't crash we lock up the system, but silently.
If we crash we lock up the system in a more obvious way.

So the decision is to make it crash :)

spacefrogg: The way I programmed it now is to not lock up the system at all but immediately return. Hyprlock has not yet told the Compositor to lock the system.

I make it crash if you want, but your statement about the current silent lock is not correct.

spacefrogg: From the perspective of the application, every case of `m_bLibFailed = true` looks identical. So, I would crash the application in all cases, then, not just when the library fails to initialise.

src/core/PwAuth.cpp
spacefrogg: I'm sorry? I just need a unique_ptr to some fixed buffer, here (for automatic deletion after the hash generation). How would you recommend doing that with `std::vector` (or, rather `std::array`)?

spacefrogg:
> > Is that not true for PAM as well? So, it should be some general artificial delay between login tries, not just for this method.
> 
> All PAM login configurations i came across have such a delay. Edit: it is just the crypto inducing this delay i think

Yeah, so I would just implement that near `hyprlock.cpp: onPasswordCheckTimer()` for all authentication interfaces.

> Finding a private key for a corresponding public key is MUCH harder than finding the plaintext for a certain hash. For that your security is only the plaintext length. Edit: if you consider the whole input range, then yes it can be considered very hard to find the plaintext of a certain hash. but by making certain assumptions about a password (charset and length), weak passwords are trivial to guess, because sha256 is fast. This is why you normally use PBKDF for password hashes.

You are right! So, it must be a read-restricted file.

PaideiaDilemma:
Since my last comment I
1. Created a PR for an authentication interface, that makes it easier to integrate a new authentication method into hyprlock. (#578)
2. Implemeneted a POC for hash based authentication using libsodium. See https://github.com/PaideiaDilemma/hyprlock/pull/2
    I think this provides a better starting point. We just need to think about how users are supposed to generate the hash, or if hyprlock even does it for you.

Issue #472, opened 2024-08-30
using a symlink to an image doesnt work


dqrk0jeste:
### Regression?

No

### Hyprlock Info and Version

<Enter your hyprlock version here>
<details> <summary>Hyprlock config</summary>

```sh
background {
  monitor =
  path = ~/image_sym
  blur_passes = 2
  contrast = 0.8916
  brightness = 0.8172
  vibrancy = 0.1696
  vibrancy_darkness = 0.0
}

general {
  ignore_empty_input = true
  no_fade_in = false
  grace = 0
  disable_loading_bar = false
}

input-field {
  monitor =
  size = 600, 60
  dots_size = 0.4
  dots_spacing = 0.2
  dots_center = true
  outer_color = rgba(0, 0, 0, 0)
  inner_color = rgba(0, 0, 0, 0)
  font_color = rgb(205, 214, 244)
  fade_on_empty = false
  placeholder_text = 
  check_color = rgba(0, 0, 0, 0)
  fail_text = 
  fail_color = rgba(0, 0, 0, 0)
  hide_input = false
  position = 0, 200
  halign = center
  valign = bottom
}

# time
label {
  monitor =
  text = cmd[update:1000] echo "<span>$(date +"%H:%M")</span>"
  color = rgba(216, 222, 233, 0.75)
  font_size = 130
  font_family = Poppins
  position = 0, 300
  halign = center
  valign = center
}

# date
label {
  monitor =
  text = cmd[update:1000] echo -e "$(date +"%A, %d %B")"
  color = rgba(216, 222, 233, 0.5)
  font_size = 30
  font_family = Poppins
  position = 0, 200
  halign = center
  valign = center
}

# current song
label {
  monitor =
  text = cmd[update:1000] echo "$(~/scripts/song.sh)" 
  color = rgba(255, 255, 255, 0.7)
  font_size = 18
  font_family = Poppins
  position = 0, 60
  halign = center
  valign = bottom
}

```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 9a09eac79b85c846e3a865a9078a3f8ff65a9259  (props: bump version to 0.42.0).
Date: Wed Aug 7 19:17:10 2024
Tag: v0.42.0, commits: 5069

flags: (if any)


System Information:
System name: Linux
Node name: archlinux
Release: 6.10.6-arch1-1
[...]

PR #576, opened 2024-12-07
refix support for symlinked images

qxb3:
Refixing #472


vaxerski (commented):
why not do it globally for all images in hyprgraphics instead? I think that'd be a better solution./

qxb3:
ah i see, the code that needs fixing has been moved to hyprgraphics, i'll close this one and make a pr there.

PR #578, opened 2024-12-08
auth: add an interface for different authentication methods

PaideiaDilemma:
This has the goal to create an interface that hyprlock uses for authentication.

After this, error messages from fingerprint can be shown in the PasswordInputField by using `$FAIL` in `input-field:fail_text`. The input field will actually show a failure when we failed to match now as well.

BREAKING:
- `general:pam_module` -> `auth:pam:module`
- `general:enable_fingerprint` -> `auth:fingerprint:enabled`
- `general:fingerprint_ready_message` -> `auth:fingerprint:ready_message`
- `general:fingerprint_present_message` -> `auth:fingerprint:present_message`

Implicit changes introduced by this PR:
- Attempts now also factor in the failed attempts from fingerprint auth.
- A fingerprint swipe clears the password input buffer.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -100,14 +100,11 @@ void CAuth::terminate() {
 }
 
 static void passwordCheckTimerCallback(std::shared_ptr<CTimer> self, void* data) {
-    const auto PCLEAR = (bool*)data;
     // check result
     if (g_pAuth->isAuthenticated()) {
         g_pHyprlock->unlock();
     } else {
-        if (*PCLEAR)
-            g_pHyprlock->clearPasswordBuffer();
-
+        g_pHyprlock->clearPasswordBuffer();
         g_pAuth->m_iFailedAttempts += 1;
         Debug::log(LOG, "Failed attempts: {}", g_pAuth->m_iFailedAttempts);
 
@@ -118,8 +115,8 @@ static void passwordCheckTimerCallback(std::shared_ptr<CTimer> self, void* data)
     }
 }
 
-void CAuth::enqueueCheckAuthenticated(bool clearPasswordBuffer) {
-    g_pHyprlock->addTimer(std::chrono::milliseconds(1), passwordCheckTimerCallback, new bool(clearPasswordBuffer));
+void CAuth::enqueueCheckAuthenticated() {
+    g_pHyprlock->addTimer(std::chrono::milliseconds(1), passwordCheckTimerCallback, nullptr);
 }
 
 void CAuth::postActivity(eAuthImplementations implType) {
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -31,15 +31,6 @@ void CAuth::submitInput(const std::string& input) {
     }
 }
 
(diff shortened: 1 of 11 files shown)


vaxerski:
> Still need to test and think about handling of fail messages from different authentication methods.

Do you want me to give you time or?

PaideiaDilemma:
Yes please, i will ping if ready

PaideiaDilemma:
@vaxerski ready.

There are some problems with input field resizing that @alba4k found that surfaced because of this PR, but unrelated. Going to fix that separately.

src/auth/Auth.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+void CAuth::terminate() {
+    for (const auto& i : m_vImpls) {
+        i->terminate();
+    }
+}
+
+static void passwordCheckTimerCallback(std::shared_ptr<CTimer> self, void* data) {
+    // check result
+    if (g_pAuth->isAuthenticated()) {
```
vaxerski, line 104: what is this {}

src/auth/Fingerprint.hpp
```diff
@@ -1,20 +1,29 @@
...
-class CFingerprint {
+class CFingerprint : public IAuthImplementation {
   public:
     CFingerprint();
 
-    std::shared_ptr<sdbus::IConnection> start();
-    bool                                isAuthenticated();
-    std::optional<std::string>          getLastMessage();
-    void                                terminate();
+    ~CFingerprint() override;
```
vaxerski, line 14: style: don't use ... override, instead use the classic virtual ...

src/auth/Auth.cpp
```diff
@@ -0,0 +1,124 @@
...
+        if (i->getImplType() == implType) {
+            return i->getLastFailText();
+        }
+    }
+    return std::nullopt;
+}
+
+std::optional<std::string> CAuth::getPrompt(eAuthImplementations implType) {
+    for (const auto& i : m_vImpls) {
+        if (i->getImplType() == implType) {
```
vaxerski, line 80: oopsie, missed these two (this one and above)

src/auth/Auth.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+static void passwordCheckTimerCallback(std::shared_ptr<CTimer> self, void* data) {
+    // check result
+    if (g_pAuth->isAuthenticated())
+        g_pHyprlock->unlock();
+    else {
+        g_pHyprlock->clearPasswordBuffer();
+        g_pAuth->m_iFailedAttempts += 1;
```
vaxerski, line 108: ++

vaxerski (commented):
looks acceptable to me

src/auth/Auth.cpp
PaideiaDilemma: a shame clang-format `RemoveBracesLLVM` removes `for` braces as well.
do you have any linter check for that?

vaxerski: nope

vaxerski (approved):

PR #582, opened 2024-12-15
input-field: fix width animations

PaideiaDilemma:
This fixes some problems with the input field resize. 
- support fading to a smaller size
- don't draw the shadow when fading
- don't draw the asset when the input field size is to small

Motivated me to start porting hyprlands animation manager to hyprlock.


vaxerski (approved):
> Motivated me to start porting hyprlands animation manager to hyprlock.

if you want to do that, (imho not a bad idea) throw it into hyprutils.

PaideiaDilemma:
@vaxerski wouldn't hyprgraphics make more sense?

vaxerski:
no?

PR #585, opened 2024-12-16
auth: add libsodium based hash authentication

PaideiaDilemma:
This is an alternative to #573 (@spacefrogg let me know what you think)

Some details:
- password generation is done via the `hyprlock-setpwhash` utility.
- salt, hashing algorithm is handled by libsodium (the salt is included in the string that we store)
- `hyprlock-setpwhash` can be followed by "interactive", "moderate" or "sensitive", to specify hashing cost. "moderate" is the default

Open stuff:
- ~~the last commit adding CSensitiveString is optional. not sure if we want it.~~ (separated out).
- `auth:sodium` or `auth:pwhash`?


PaideiaDilemma:
abandoning for now : )

PR #590, opened 2024-12-19
misc: add git commit hash to --version when not on tag v${VERSION}

PaideiaDilemma:
Opted for doing it straight in cmake, since we don't need all the stuff that `generateVersion.sh` in Hyprland does.
This only appends the short git hash when we are not on the tag v${VERSION} where ${VERSION} is the content of the VERSION file.

Merging this implies that `Hyprlock version 0.6.0` can be actual v0.6.0, https://github.com/hyprwm/hyprlock/commit/058830668e92e92b3179319090b5d523ba4d59a4 or https://github.com/hyprwm/hyprlock/commit/d12b4a7fbaf570fbe9322c09a777c124fbf15306, since those 2 commit don't have this yet

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,9 +28,28 @@ set(CMAKE_CXX_STANDARD 23)
 add_compile_options(-Wall -Wextra -Wno-unused-parameter -Wno-unused-value
                     -Wno-missing-field-initializers -Wno-narrowing)
 
+# get git commit
+execute_process(
+  OUTPUT_VARIABLE GIT_SHORT_HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND git rev-parse --short HEAD
+)
+
+# get git commit of v$VERSION
+execute_process(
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -28,9 +28,30 @@ set(CMAKE_CXX_STANDARD 23)
...
+  OUTPUT_VARIABLE GIT_TAG_SHORT_HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND git rev-parse --short --verify v${VERSION}
+)
+
+if(GIT_SHORT_HASH STREQUAL GIT_TAG_SHORT_HASH)
+  add_compile_definitions(HYPRLOCK_VERSION="${VERSION}")
+else()
+  add_compile_definitions(HYPRLOCK_VERSION="${VERSION}+${GIT_SHORT_HASH}")
```
vaxerski, line 50: I'd rather not do this and instead format at runtime as needed

PaideiaDilemma: smth like this?

vaxerski (approved):
ye thanks

PR #591, opened 2024-12-19
config: add sizex using CLayoutXValueData for reproducible layouting of image and label

### LINKED ISSUE #261: [Feat Req ] Follow Scaling per monitor by default
Maybe I missed something. For QOL, is it possible for the lockscreen layer to follow per monitor scaling by default?
 It's possible to script it and generate compatible per monitor configs, but would be nice to have this by default. Thank you! 

Example
*  mon1 has scaling of 2 while mon2 has 1.  
* hidpi monitors together with non hidpi
* Easy to reproduced for other users 😉

### LINKED ISSUE #555: Add the ability to set the image size as a percentage
### Description

Currently, only shape size can be set by a percentage and image sizes can't.

PaideiaDilemma:
Closes #555
Closes #261

BREAKING: removed `image:size` in favour of `image:sizex`.

**Image**:

Image is now dimensioned via `sizex` which allows you to specify the x dimension in px or %. The y dimension is inferred from the ratio of the image.

**Label**:

Label keeps the `font_size` option and `sizex` will be 0 per default. As long as `sizex` stays 0, labels behave the same as currently.
However, you can now set the absolute size of a label with `sizex` (y is inferred from the texture). In combination with `font_size` this allows to configure the _sharpness_ in a way.

### DIFF
--- a/src/config/ConfigDataValues.hpp
+++ b/src/config/ConfigDataValues.hpp
@@ -12,8 +12,9 @@ using namespace Hyprutils::String;
 
 enum eConfigValueDataTypes {
     CVD_TYPE_INVALID  = -1,
-    CVD_TYPE_LAYOUT   = 0,
+    CVD_TYPE_LAYOUTXY = 0,
     CVD_TYPE_GRADIENT = 1,
+    CVD_TYPE_LAYOUTX  = 2,
 };
 
 class ICustomConfigValueData {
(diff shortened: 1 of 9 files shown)


PaideiaDilemma:
Not sure about the breaking change for images.
I don't like that it is currently called size, since that option is 2D everywhere else and you also don't really specify the size currently. Its more like a scaling option.

PaideiaDilemma:
Wiki mr:  hyprwm/hyprland-wiki#912

Memoraike:
Really looking forward to this MR :)
<details><summary>Details</summary>
<p>

image

image

</p>
</details>

PR #604, opened 2024-12-23
auth: assert username

PaideiaDilemma:
Note: if this assert fires, it will be before acquiring the session lock.


vaxerski (approved):
makes sense

PR #609, opened 2024-12-25
Add D-Bus signals to notify session lock and unlock states

Iamanaws:
Send an "Unlock" D-Bus signal to notify programs listening for Unlock events.

This enables tools like hypridle to detect the Unlock event and execute the unlock_cmd.

Issues:
hyprwm/hypridle#79
hyprwm/hypridle#112

### DIFF
--- a/src/core/DBusManager.cpp
+++ b/src/core/DBusManager.cpp
@@ -1,0 +1,88 @@
+#include "DBusManager.hpp"
+#include "../helpers/Log.hpp"
+
+DBusManager& DBusManager::getInstance() {
+    static DBusManager instance;
+    return instance;
+}
+
+DBusManager::DBusManager() {
+    initializeConnection();
+}
+
+DBusManager::~DBusManager() {
+    // Resources are automatically cleaned up.
+}
+
+void DBusManager::initializeConnection() {
+    try {
+        m_connection = sdbus::createSystemBusConnection();
+
+        const sdbus::ServiceName destination{"org.freedesktop.login1"};
+        const sdbus::ObjectPath loginPath{"/org/freedesktop/login1"};
+        const sdbus::ObjectPath sessionPath{"/org/freedesktop/login1/session/auto"};
+
+        m_loginProxy = sdbus::createProxy(*m_connection, destination, loginPath);
+        m_sessionProxy = sdbus::createProxy(*m_connection, destination, sessionPath);
+
+        Debug::log(LOG, "[DBusManager] Initialized D-Bus connection. Service: {}. Login path: {}, Session path: {}",
+            std::string(destination), std::string(loginPath), std::string(sessionPath));
+    } catch (const sdbus::Error& e) {
+        Debug::log(ERR, "[DBusManager] D-Bus connection initialization failed: {}", e.what());
+    }
+}
+
+std::shared_ptr<sdbus::IConnection> DBusManager::getConnection() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_connection;
+}
+
+std::shared_ptr<sdbus::IProxy> DBusManager::getLoginProxy() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (!m_loginProxy) {
+        initializeConnection();
+    }
+    return m_loginProxy;
+}
+
+std::shared_ptr<sdbus::IProxy> DBusManager::getSessionProxy() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (!m_sessionProxy) {
+        initializeConnection();
+    }
+    return m_sessionProxy;
+}
+
+void DBusManager::setLockedHint(bool locked) {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (!m_sessionProxy) {
+        Debug::log(WARN, "[DBusManager] Cannot set locked hint: Proxy is not initialized.");
+        return;
+    }
+
+    try {
+        const sdbus::ServiceName interface{"org.freedesktop.login1.Session"};
+        m_sessionProxy->callMethod("SetLockedHint").onInterface(interface).withArguments(locked);
+
+        Debug::log(LOG, "[DBusManager] Sent 'SetLockedHint({})' on {}", locked, std::string(interface));
+    } catch (const sdbus::Error& e) {
+        Debug::log(WARN, "[DBusManager] Failed to send 'SetLockedHint({})': {}", locked, e.what());
+    }
+}
+
+void DBusManager::sendUnlockSignal() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (!m_sessionProxy) {
+        Debug::log(WARN, "[DBusManager] Unlock signal skipped: Proxy is not initialized.");
+        return;
+    }
+
+    try {
+        const sdbus::ServiceName interface{"org.freedesktop.login1.Session"};
+        m_sessionProxy->callMethod("Unlock").onInterface(interface);
+
+        Debug::log(LOG, "[DBusManager] Sent 'Unlock' on {}", std::string(interface));
+    } catch (const sdbus::Error& e) {
+        Debug::log(WARN, "[DBusManager] Unlock signal failed: {}", e.what());
+    }
+}
--- a/src/core/DBusManager.cpp
+++ b/src/core/DBusManager.cpp
@@ -33,12 +33,10 @@ void DBusManager::initializeConnection() {
 }
 
 std::shared_ptr<sdbus::IConnection> DBusManager::getConnection() {
-    std::lock_guard<std::mutex> lock(m_mutex);
     return m_pConnection;
 }
 
 std::shared_ptr<sdbus::IProxy> DBusManager::getLoginProxy() {
-    std::lock_guard<std::mutex> lock(m_mutex);
     if (!m_pLoginProxy)
         initializeConnection();
 
@@ -46,7 +44,6 @@ std::shared_ptr<sdbus::IProxy> DBusManager::getLoginProxy() {
 }
 
 std::shared_ptr<sdbus::IProxy> DBusManager::getSessionProxy() {
-    std::lock_guard<std::mutex> lock(m_mutex);
     if (!m_pSessionProxy)
         initializeConnection();
 
@@ -54,7 +51,6 @@ std::shared_ptr<sdbus::IProxy> DBusManager::getSessionProxy() {
 }
 
 void DBusManager::setLockedHint(bool locked) {
-    std::lock_guard<std::mutex> lock(m_mutex);
     if (!m_pSessionProxy) {
         Debug::log(WARN, "[DBusManager] Cannot set locked hint: Proxy is not initialized.");
         return;
@@ -71,7 +67,6 @@ void DBusManager::setLockedHint(bool locked) {
 }
 
 void DBusManager::sendUnlockSignal() {
-    std::lock_guard<std::mutex> lock(m_mutex);
     if (!m_pSessionProxy) {
         Debug::log(WARN, "[DBusManager] Unlock signal skipped: Proxy is not initialized.");
         return;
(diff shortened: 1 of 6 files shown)


PaideiaDilemma:
While i understand what you want to achieve, i think that is not the way this dbus interface is supposed to work.

From [org.freedesktop.login1](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html):

> Lock()/Unlock() is sent when the session is asked to be screen-locked/unlocked. A session manager of the session should listen to this signal and act accordingly. This signal is sent out as a result of the Lock() and Unlock() methods, respectively.

So from what i understand `Unlock` is not supposed to signal that we just unlocked, but to request an unlock?? Not entirely clear to me how it is supposed to be used though.

Iamanaws:
Yes, I agree it’s a bit unclear, but from what I understand, Unlock() is intended to inform the Session Manager and other programs listening to D-Bus that the lock screen should be removed. It’s up to each listening process to handle this notification as they see fit.

In this case, Hyprlock directly manages the unlock process via Wayland using the ext_session_lock_v1 interface. In other scenarios, a lock screen might call Unlock() to signal the Session Manager that the unlock should proceed after successful authentication.

The key point is that Unlock() does not handle the unlocking itself—it simply notifies that it should occur or has occurred. By sending the Unlock() method after handling the unlock, Hyprlock ensures compatibility with tools and listeners on D-Bus, even if it doesn’t rely on them.

Since Hyprlock already performs the actual unlock via Wayland, calling Unlock() is purely informational and has no functional impact on the unlocking process.

PaideiaDilemma:
Can't you just use smth like `hyprlock && loginctl unlock-session`??

PaideiaDilemma:
But i agree that it would be better if hypridle's unlock_cmd just worked...
Do you know any lock screen app that sends Unlock?

Iamanaws:
> Can't you just use smth like `hyprlock && loginctl unlock-session`??

Yes, but we should map every hyprlock call or make an alias.


> But i agree that it would be better if hypridle's unlock_cmd just worked... Do you know any lock screen app that sends Unlock?

I'm looking into it, there's not much documentation or code about it out there. 

I see GDM calls login1.Manager UnlockSession() method that uses Session Unlock. [GDM](https://gitlab.gnome.org/GNOME/gdm/-/blob/main/daemon/gdm-manager.c#L297-327)

> LockSession() asks the session with the specified ID to activate the screen lock. UnlockSession() asks the session with the specified ID to remove an active screen lock, if there is any. This is implemented by sending out the Lock() and Unlock() signals from the respective session object which session managers are supposed to listen on. [org.freedesktop.login1.Manager#Methods](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html#Methods)

Iamanaws:
Also I think for `unlock_cmd` we could create a custom Hyprlock interface that sends a custom signal with Unlock as member because Hypridle is not match the login1.Session Interface, just the member of the interface should match the name "Unlock".

That could work if you are concern about using, how to and the repercussion of calling `Session.Unlock()`

Iamanaws:
i3-lock and swaylock lack the functionality.

But I found there's a `LockedHint` property that is part of the Session too and could be an option. 
I like it because it removes the ambiguity of Unlock() signal, but in that case, I will have to make a change to Hypridle to be listening to the LockedHint property, and will have to test how well it works.

Also, I found a comment with an interesting interpretation of the [org.freedesktop.login1](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html) documentation and how it could be properly implemented.

[](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/741#note_2288398)

PaideiaDilemma:
`LockedHint` sounds like a better solution thanks!

The only thing is that we now would have 2 dbus connection objects for org.freedesktop.login1, when a user enables fingerprint authentication as well.

We should centralize that by implementing a dbus manager class. Idk if you would want to try to do that. If you don't , I think we can do it afterwards.

Also it would be good to add an option to enable/disable this. Enable per default.

Iamanaws:
I centralized the dbus connection, could you give it a look please?

src/core/hyprlock.hpp
```diff
@@ -12,6 +12,7 @@
 #include <vector>
 #include <condition_variable>
 #include <optional>
+#include <sdbus-c++/sdbus-c++.h>
```
PaideiaDilemma, line 15: i think you can move the header back to the source file right?

src/core/DBusManager.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+std::shared_ptr<sdbus::IConnection> DBusManager::getConnection() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    return m_connection;
+}
+
+std::shared_ptr<sdbus::IProxy> DBusManager::getLoginProxy() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (!m_loginProxy) {
```
PaideiaDilemma, line 42: style: no baces for if with a single statement. same below

src/core/DBusManager.hpp
```diff
@@ -0,0 +1,30 @@
...
+    void setLockedHint(bool locked);
+    void sendUnlockSignal();
+
+private:
+    DBusManager();
+    ~DBusManager();
+
+    void initializeConnection();
+
+    std::shared_ptr<sdbus::IConnection> m_connection;
```
PaideiaDilemma, line 25: style: call em `m_pConnection`, `m_pLoginProxy` and `m_pSessionProxy`

src/core/DBusManager.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        Debug::log(LOG, "[DBusManager] Initialized D-Bus connection. Service: {}. Login path: {}, Session path: {}",
+            std::string(destination), std::string(loginPath), std::string(sessionPath));
+    } catch (const sdbus::Error& e) {
+        Debug::log(ERR, "[DBusManager] D-Bus connection initialization failed: {}", e.what());
+    }
+}
+
+std::shared_ptr<sdbus::IConnection> DBusManager::getConnection() {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
PaideiaDilemma, line 36: this lock does nothing cause it passes its raw resource out of the function, same below. returning the pointer itself is atomic anyways.

src/core/DBusManager.cpp
Iamanaws: https://github.com/hyprwm/hyprlock/pull/609/commits/00b02c005f0d623c66ff15b0167d04a2859e3eae

src/core/DBusManager.hpp
Iamanaws: https://github.com/hyprwm/hyprlock/pull/609/commits/00b02c005f0d623c66ff15b0167d04a2859e3eae

src/core/DBusManager.cpp
Iamanaws: Got it, thanks for the feedback! 
I removed the locks (https://github.com/hyprwm/hyprlock/pull/609/commits/b22009f992124953e4e1fdd5b30750d04b803c15), and since our usage is pretty low-volume I don't think we really need multiple threads at the moment.

PaideiaDilemma: Right we don't even use the proxies in different threads.
In that case can you use `SP<>` (shared pointer impl from hyprutils) instead of `std::shared_ptr`?

PaideiaDilemma:
I am still a bit unsure if this is the right solution.

For me it is unclear if it is appropriate to call this method as a session-lock client.
I think it would be more appropriate for the compositor that implements the server side of session-lock to set the locked hint. But hyprland currently doesn't do dbus stuff and i think that is fine. Now if some compositor would implement it, I guess we would not break it, we would just set the hint twice right?.

What i also did not consider initially is that `unlock_cmd` actually has a proper functionality already, but is easy to be misunderstood. You could for example do
```init
unlock_cmd = pkill -USR1 hyprlock
``` 
to have `loginctl unlock-session` unlock the session properly.
this would not make any sense any more, if unlock_cmd would be triggered by hyprlock!

However signaling lock/unlock is also related to the start before suspend problem that we also have. If we can properly signal lock and unlock, we can solve that by waiting for locked in before_sleep_cmd.

What do you think about just implementing an internal way of signaling lock/unlock to hypridle and use that for two additional triggers in hypridle and to wait for locked when launching hyprlock via before_sleep_cmd.

Iamanaws:
> For me it is unclear if it is appropriate to call this method as a session-lock client. I think it would be more appropriate for the compositor that implements the server side of session-lock to set the locked hint. But hyprland currently doesn't do dbus stuff and i think that is fine. Now if some compositor would implement it, I guess we would not break it, we would just set the hint twice right?.

Yeah, also though about the compositor handling it at first, but it seems like a simpler solution to just do it here. And yes, I think there should be no problem if it is set twice.


> What i also did not consider initially is that `unlock_cmd` actually has a proper functionality already, but is easy to be misunderstood. You could for example do
> 
> ```
> unlock_cmd = pkill -USR1 hyprlock
> ```
> 
> to have `loginctl unlock-session` unlock the session properly. this would not make any sense any more, if unlock_cmd would be triggered by hyprlock!

That's a bit confusing. unlock_cmd would never be called because hyprlock does not tell hypridle that it unlocked / not send dbus signal. But yeah, when it does it will be redundant.

***
> What do you think about just implementing an internal way of signaling lock/unlock to hypridle and use that for two additional triggers in hypridle and to wait for locked when launching hyprlock via before_sleep_cmd.

I like the idea of an internal unlock signal. However, I still see value in keeping the D-Bus signals for broader compatibility, but not a big problem.
I think we can maintain Hypridle’s current behavior (listening for unlock signals) while also triggering unlock_cmd upon receiving an internal signal. That signal could originate from Hyprlock, or even better, from Hyprland. This approach would also support other lock screens that don’t use D-Bus. If keeping the D-Bus listener on Hypridle becomes redundant, we can remove it later.

I'm not sure if there's already, and if there is how, a channel used by Hyprland and Hypridle or hypr apps in general to communicate between them, could you point me a to? I think there is, just not very familiar with.

PaideiaDilemma:
>  That signal could originate from Hyprlock, or even better, from Hyprland. This approach would also support other lock screens that don’t use D-Bus.

I think that is a really good idea! Have Hyprland signal to hypridle whether we are locked or not.

> I'm not sure if there's already, and if there is how, a channel used by Hyprland and Hypridle or hypr apps in general to communicate between them, could you point me to? I think there is, just not very familiar with.

Well the existing channel is wayland. I don't think there is anything else that we could use for that purpose. We would need an internal protocol for that. It could work exactly like the idle-notify protocol. As a wayland client we can request to be notified and then we will receive a `locked` and `unlock` from the compositor.
I will ask vaxry what he thinks about that.

davidmjerry:
> Can't you just use smth like `hyprlock && loginctl unlock-session`??

This seems to just send the second command immediately after locking

PaideiaDilemma:
@vaxerski 

What do you think about this idea?

We are trying to solve the "lock before suspend problem" as well as giving hyprIDLE the ability to launch commands when we unlock.

swaylock has `-R, --ready-fd` to script your way out of the suspend problem.
Instead of adding something like that (See https://github.com/hyprwm/hyprlock/pull/637), we could solve this problem on the compositor level, by adding a hypothetical "hyprland-session-lock-notify" protocol that would work very similar to "ext-idle-notfiy-v1".
Then in hypridle we would be able to receive lock and unlock notifications allowing it to
1.  inihibit sleep via dbus until we are actually locked.
2.  launch a command on unlock.

vaxerski (commented):
also in dire need of a clang-format

PaideiaDilemma:
@Iamanaws vaxry misunderstood me and reviewed the pr instead of the comment, but he has nothing against such a protocol. I will make one in hyprland-protocols and start a hyprland implementation : )

PaideiaDilemma:
@Iamanaws I created https://github.com/hyprwm/hyprland-protocols/pull/12 and https://github.com/hyprwm/Hyprland/pull/9092

You wanna do the hypridle impl, or should i?

Iamanaws:
@PaideiaDilemma go for it, you're half the way

Iamanaws:
https://github.com/hyprwm/hypridle/pull/122

Issue #608, opened 2024-12-24
Conflict between parallel password and fingerprint authentication


rickyelopez:
### Regression?

No

### Hyprlock Info and Version

Hyprlock version 0.6.0
<details> <summary>Hyprlock config</summary>

```sh
general {
  grace = 1
  no_fade_in = false
  no_fade_out = false
}

auth {
  fingerprint {
    enabled = true
  }
}

background {
  monitor = 
  path = screenshot
  color = rgba(25, 20, 20, 1.0)
  blur_passes = 2
}

input-field {
  monitor =
  size = 200, 50
  outline_thickness = 3
  dots_size = 0.11
  fade_timeout = 2000

  position = 0, -300
  halign = center
  valign = center
}

label {
  monitor =
  text = $TIME
  color = rgb(100, 100, 100)
  font_size = 50
  font_family = IBM Plex

  position = 0, 250
  halign = center
  valign = center
}
```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.46.2 built from branch  at commit 0bd541f2fd902dbfa04c3ea2ccf679395e316887  (version: bump to 0.46.2).
Date: Thu Dec 19 19:26:47 2024
Tag: v0.46.2, commits: 5566
built against:
 aquamarine 0.5.1
 hyprlang 0.6.0
 hyprutils 0.2.6
 hyprcursor 0.1.10
 hyprgraphics 0.1.1


flags set:
debug


System Information:
System name: Linux
Node name: expedition
Release: 6.12.4-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Mon, 09 Dec 2024 14:31:57 +0000


GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation CometLake-H GT2 [UHD Graphics] [8086:9bc4] (rev 05) (prog-if 00 [VGA controller])
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation TU117M [GeForce GTX 1650 Mobile / Max-Q] [10de:1f99] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  565.77  Wed Nov 27 23:33:08 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
[...]

PR #613, opened 2024-12-25
Better feedback from fingerprint auth and don't clear password field on retry (only on failure)

rickyelopez:
Quick and crude fix targetted at the issue I raised in #608.

### DIFF
--- a/src/auth/Fingerprint.cpp
+++ b/src/auth/Fingerprint.cpp
@@ -13,7 +13,6 @@ static const auto FPRINT        = sdbus::ServiceName{"net.reactivated.Fprint"};
 static const auto DEVICE        = sdbus::ServiceName{"net.reactivated.Fprint.Device"};
 static const auto MANAGER       = sdbus::ServiceName{"net.reactivated.Fprint.Manager"};
 static const auto LOGIN_MANAGER = sdbus::ServiceName{"org.freedesktop.login1.Manager"};
-static const auto RETRY_MESSAGE = "Could not match fingerprint. Try again.";
 
 enum MatchResult {
     MATCH_INVALID = 0,
@@ -81,10 +80,14 @@ void CFingerprint::handleInput(const std::string& input) {
 }
 
 std::optional<std::string> CFingerprint::getLastFailText() {
-    return m_sDBUSState.message.empty() ? std::nullopt : std::optional(m_sDBUSState.message);
+    if (!m_sFailureReason.empty())
+        return std::optional(m_sFailureReason);
+    return std::nullopt;
 }
 
 std::optional<std::string> CFingerprint::getLastPrompt() {
+    if (!m_sPrompt.empty())
+        return std::optional(m_sPrompt);
     return std::nullopt;
 }
 
@@ -140,7 +143,8 @@ bool CFingerprint::createDeviceProxy() {
                 bool       isPresent      = presentVariant.get<bool>();
                 if (!isPresent)
                     return;
-                m_sDBUSState.message = m_sFingerprintPresent;
+                m_sPrompt        = m_sFingerprintPresent;
+                m_sFailureReason = "";
                 g_pHyprlock->enqueueForceUpdateTimers();
             } catch (std::out_of_range& e) {}
         });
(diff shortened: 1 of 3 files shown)


src/auth/Fingerprint.hpp
```diff
@@ -27,6 +27,8 @@ class CFingerprint : public IAuthImplementation {
   private:
     struct SDBUSState {
         std::string                         message = "";
+        std::string                         failure_reason = "";
```
PaideiaDilemma, line 30: `failureReason`

rickyelopez: oh I also forgot to delete `message` after refactoring

PaideiaDilemma: Actually while your at it, can you move the strings out of `SDBUSState`?
Either just `m_sMessage, m_sFailureState` and co, or in their own struct. Feel like they don't belong there.

rickyelopez: sure that's fair

src/auth/Fingerprint.cpp
```diff
@@ -81,10 +81,14 @@ void CFingerprint::handleInput(const std::string& input) {
 }
 
 std::optional<std::string> CFingerprint::getLastFailText() {
-    return m_sDBUSState.message.empty() ? std::nullopt : std::optional(m_sDBUSState.message);
+    if (!m_sFailureReason.empty() && m_sPrompt.empty())
```
PaideiaDilemma, line 84: no need to check for prompt i think. same below

src/auth/Fingerprint.cpp
```diff
@@ -153,41 +158,61 @@ void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
...
             stopVerify();
-            m_sDBUSState.message = "Fingerprint auth disabled (unknown error)";
+            m_sPrompt         = "";
+            m_sFailureReason = "Fingerprint auth disabled (unknown error)";
             break;
         case MATCH_MATCHED:
             stopVerify();
-            m_sDBUSState.message = "";
-            authenticated        = true;
+            m_sPrompt         = "Fingerprint matched!";
```
PaideiaDilemma, line 183: Wait that is not a "prompt"?
I prompt should indicate what we expect the user to do. `Password: `, `TOTP: ` or in this context smth like `Put your finger on the sensor`.

src/auth/Fingerprint.cpp
rickyelopez: yeah you're right, originally I wasn't clearing one when setting the other, but I think now it should work the same without this check

src/auth/Fingerprint.cpp
rickyelopez: fair enough

src/auth/Fingerprint.cpp
```diff
@@ -218,15 +243,20 @@ void CFingerprint::startVerify(bool isRetry) {
...
+                m_sFailureReason = "Fingerprint auth disabled (failed to restart)";
+            }
         } else {
             Debug::log(LOG, "fprint: started verifying");
             if (isRetry) {
                 m_sDBUSState.retries++;
-                m_sDBUSState.message = RETRY_MESSAGE;
-            } else
-                m_sDBUSState.message = m_sFingerprintReady;
+                m_sPrompt        = RETRY_MESSAGE;
```
rickyelopez, line 254: @PaideiaDilemma just curious, why is this the only message that is predefined as a const?

PaideiaDilemma: No particular reason, and its not in style as well (Consts are without `_`). You could remove it and substitute for the string.

PaideiaDilemma (approved):

PaideiaDilemma:
@rickyelopez needs a wiki mr for `$FPRINTPROMPT` :)

rickyelopez:
> @rickyelopez needs a wiki mr for `$FPRINTPROMPT` :)

Just put up a PR: https://github.com/hyprwm/hyprland-wiki/pull/915

vaxerski (approved):

PR #625, opened 2024-12-29
fingerprint: add a delay after an unrecognized fingerprint

BlueZeeKing:
Add a new config option ("auth:fingerprint:retry_delay") that adds a delay after an unrecognized fingerprint has been scanned. The default delay is 250ms. Currently when an incorrect finger is scanned, it gets scanned as fast as possible which leads to the number of retries being using up extremely quickly.

### DIFF
--- a/src/auth/Fingerprint.cpp
+++ b/src/auth/Fingerprint.cpp
@@ -169,7 +169,13 @@ void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
                 m_sFailureReason = "Fingerprint auth disabled (too many failed attempts)";
             } else {
                 done = false;
-                startVerify(true);
+                static const auto RETRYDELAY = **(Hyprlang::INT* const*)(g_pConfigManager->getValuePtr("auth:fingerprint:retry_delay"));
+                g_pHyprlock->addTimer(
+                    std::chrono::milliseconds(RETRYDELAY),
+                    [](std::shared_ptr<CTimer> self, void* data) {
+                        ((CFingerprint*)data)->startVerify(true);
+                    },
+                    this);
             }
             break;
         case MATCH_UNKNOWN_ERROR:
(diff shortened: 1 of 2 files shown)


src/auth/Fingerprint.cpp
```diff
@@ -169,7 +169,14 @@ void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
                 m_sFailureReason = "Fingerprint auth disabled (too many failed attempts)";
             } else {
                 done = false;
-                startVerify(true);
+                g_pHyprlock->addTimer(
+                    std::chrono::milliseconds(**(Hyprlang::INT* const*)(g_pConfigManager->getValuePtr("auth:fingerprint:retry_delay"))),
```
PaideiaDilemma, line 173: can you move this out of the argument to `addTimer`??
```cpp
static auto* const PRETRYDELAY = **(Hyprlang::INT* const*)g_pConfigManager->getValuePtr("auth:fingerprint:retry_delay")
```

src/auth/Fingerprint.cpp
```diff
@@ -169,7 +169,14 @@ void CFingerprint::handleVerifyStatus(const std::string& result, bool done) {
...
+                    [](std::shared_ptr<CTimer> self, void* data) {
+                        ((CFingerprint*)data)->startVerify(true);
+                        g_pAuth->m_bDisplayFailText = false;
```
PaideiaDilemma, line 176: Why do you to set this to false?

PaideiaDilemma (changes requested):

src/auth/Fingerprint.cpp
BlueZeeKing: I'm honestly not sure. I think I was testing something and forgot to remove it.

PaideiaDilemma (approved):

PR #637, opened 2025-01-03
allow signaling a parent process that the screen was locked

### LINKED ISSUE #633: Desktop briefly flashes and gets focused before hyprlock is opened upon the system moving out of suspend
When I open my laptop lid, I briefly see my desktop, and then hyprlock shows up. This is with a bone stock hypridle config (the code from the example hypridle config on the hyprland wiki), even worse, hyprlock isn't focused, so unless I move my mouse, I type in other applications.

I shouldn't have to state why this is an absolutely horrible security hole, but more importantly for me, it's just plain annoying.

here's my hypridle config:
```
general {
    lock_cmd = pidof hyprlock || hyprlock       # avoid starting multiple hyprlock instances.
    before_sleep_cmd = loginctl lock-session    # lock before suspend.
    after_sleep_cmd = hyprctl dispatch dpms on  # to avoid having to press a key twice to turn on the display.
}

listener {
    timeout = 150                                # 2.5min.
    on-timeout = brightnessctl -s set 10         # set monitor backlight to minimum, avoid 0 on OLED monitor.
    on-resume = brightnessctl -r                 # monitor backlight restore.
}

listener {
    timeout = 300                                 # 5min
    on-timeout = loginctl lock-session            # lock screen when timeout has passed
}

listener {
    timeout = 330                                 # 5.5min
    on-timeout = hyprctl dispatch dpms off        # screen off when timeout has passed
    on-resume = hyprctl dispatch dpms on          # screen on when activity is detected after timeout has fired.
}

listener {
    timeout = 1200                                # 20min
    on-timeout = systemctl suspend                # suspend pc
}
```

I've tried futzing around with the `before_sleep_cmd` in my hypridle config by changing it to activate `hyprlock --immediate` to no avail, as well as adding lines to my `hyprland.conf` that either run `systemctl suspend` or `hyprlock --immediate` on laptop lid close (and the latter of which I also tried on laptop lid open. But when I ran hyprlock on both lid close and lid open, hyprlock was frozen entirely)

Any help/advice? I really love hyprlock because of how damn good looking it is, I really want it to work.

dkondor:
(partially) fixes #633 

Hi!
I've recently tried out hyprlock and it works nicely with my setup! (using Wayfire mostly) However, I also noticed the issue reported in #633 and I think it is mainly caused by not being able to delay suspend until the screen is locked.

This patch adds two (related) ways to signal to a parent process when the screen has been locked:

1. Writing to an inherited file descriptor: after the screen has been locked, a single newline is written to the FD given as an argument to the `-R` / `--ready-fd` option. Similar functionality exists in [swaylock](https://github.com/swaywm/swaylock/blob/master/swaylock.1.scd).
2. Creating a child process (with `fork()`) and waiting until it locks the screen (`-D` / `--daemonize` option). This ensures that when the parent process exits, the screen is locked. This is equivalent to the `-fork-on-lock` option in [waylock](https://codeberg.org/ifreund/waylock/src/branch/master/doc/waylock.1.scd) or the `-f` / `--daemonize` option of swaylock.

Using the second option and e.g. [swayidle](https://github.com/swaywm/swayidle), it can be ensured that the screen is locked before suspend:
```
swayidle -w before-sleep "hyprlock -D"
```
The first option can be tested e.g. with the screenlocker proxy service I'm working on:
https://github.com/dkondor/wayland-screenlock-proxy/tree/hyprlock

Caveats / TODO:
 - Not tested with hypridle yet (but likely would need the equivalent of the `-w` option in swayidle unless I'm missing something)
 - This only works well if  `no_fade_in = true` is set in the config, since it signals success immediately when getting the [locked](https://wayland.app/protocols/ext-session-lock-v1#ext_session_lock_v1:event:locked) event. I looked into how the fade-in animation is managed, and seems I would need to check when `bga == 1` [here](https://github.com/hyprwm/hyprlock/blob/e01afaf107650eb7f84ffe6444d1cfcfbb4a0ff5/src/renderer/Renderer.cpp#L235), but I'm unsure how to get this information to the `CHyprlock` class yet.
 - Both options do not work well together (if both `-D` and `-R` is given, the latter is ignored); not sure if this should be an error, or the worked around to still signal the fd given to `-R`

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -12,6 +12,7 @@
 #include <fcntl.h>
 #include <signal.h>
 #include <unistd.h>
+#include <errno.h>
 #include <assert.h>
 #include <string.h>
 #include <xf86drm.h>
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -23,7 +23,9 @@
 
 using namespace Hyprutils::OS;
 
-CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const bool noFadeIn) {
+CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const bool noFadeIn, const int notifyFd) {
+    m_iNotifyFd = notifyFd;
+
     m_sWaylandState.display = wl_display_connect(wlDisplay.empty() ? nullptr : wlDisplay.c_str());
     if (!m_sWaylandState.display) {
         Debug::log(CRIT, "Couldn't connect to a wayland compositor");
@@ -1013,6 +1015,16 @@ void CHyprlock::onLockLocked() {
     Debug::log(LOG, "onLockLocked called");
 
     m_bLocked = true;
+
+    if (m_iNotifyFd >= 0) {
+        const char ready = '\n';
+        while (true) {
+            if (write(m_iNotifyFd, &ready, 1) == 1 || errno != EINTR)
+                break;
+        }
+        close(m_iNotifyFd);
+        m_iNotifyFd = -1;
+    }
 }
 
 void CHyprlock::onLockFinished() {
--- a/src/core/hyprlock.hpp
+++ b/src/core/hyprlock.hpp
@@ -26,7 +26,7 @@ struct SDMABUFModifier {
 
 class CHyprlock {
   public:
-    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const bool noFadeIn);
+    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const bool noFadeIn, const int notifyFd);
     ~CHyprlock();
 
     void                            run();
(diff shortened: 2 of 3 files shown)


PaideiaDilemma:
I think vaxry and me probably both thought about adding this.
Swaylock does that, but I thinks it is not an ideal solutions, since you still need to script it yourself.

PaideiaDilemma:
Check out https://github.com/hyprwm/hypridle/pull/122 as well : )

dkondor:
> Check out [hyprwm/hypridle#122](https://github.com/hyprwm/hypridle/pull/122) as well : )

I was also looking at #609 that can be another alternative by setting `org.freedesktop.login1.Session.LockedHint` (although I wonder if it should set it to true later, after displaying the lock surface).
I'll close this if either of them get merged.

PaideiaDilemma:
@dkondor https://github.com/hyprwm/hypridle/pull/122 has been merged. Let me know if you have any feedback.

dkondor:
> @dkondor [hyprwm/hypridle#122](https://github.com/hyprwm/hypridle/pull/122) has been merged. Let me know if you have any feedback.

Will test and do so!

PR #641, opened 2025-01-06
auth: fixup prompt and fail substitution

PaideiaDilemma:
This PR brings back fingerprint errors in `$FAIL`. That was regressed by https://github.com/hyprwm/hyprlock/pull/603.

BREAKING:
- Removed $PROMPT variable. Either use $PAMPROMPT or $FPRINTPROMPT.
- Removed $FPRINTMESSAGE. Use $FPRINTPROMPT instead. There is also
  $FPRINTFAIL.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -40,21 +40,8 @@ bool CAuth::checkWaiting() {
     return false;
 }
 
-std::string CAuth::getInlineFeedback() {
-    std::optional<std::string> firstFeedback = std::nullopt;
-    for (const auto& i : m_vImpls) {
-        const auto FEEDBACK = (m_bDisplayFailText) ? i->getLastFailText() : i->getLastPrompt();
-        if (!FEEDBACK.has_value())
-            continue;
-
-        if (!firstFeedback.has_value())
-            firstFeedback = FEEDBACK;
-
-        if (i->getImplType() == m_eLastActiveImpl)
-            return FEEDBACK.value();
-    }
-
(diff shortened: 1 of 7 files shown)


PaideiaDilemma:
After pondering with @alba4k we thought removing `PROMPT` was appropriate, since it is not that useful and we don't need to solve the "when to show which auth prompt" problem.

$FAIL and `g_pAuth->getLastFailText()` are now handled by the auth method directly (they specify the failText via `g_pAuth->enqeueFail`.

PaideiaDilemma:
@vaxerski this one is ready.

Regarding the breaking variable changes (PROMPT & FPRINTMESSAGE), i could instead:
- not remove them yet and say they will be removed in a future release in the wiki
- replace them with a message to the user. smth like `$PROMPT is Depreciated!`

Let me know if you think any of those make more sense.
Personally i think not many people use those vars, so i think it is not that big of a problem.

alba4k:
I think they can be safely be removed, but if you really want to be safe the easiest solution is probably to just keep them working for a while and print a warning to stderr saying that the variable is deprecated and will be removed soon.

vaxerski:
why do the breaking changes? is there any good reason?

alba4k:
FPRINTMESSAGE is a misleading name, FPRINTFAIL is more consistent with the other variables and explains the purpose of the variable better

a unified PROMT variable doesn't really make sense with more than one auth method. Separating the variables makes more sense and gives the user more control

vaxerski (approved):
thanks. Wiki mr?

vaxerski:
oops, already there.

PR #642, opened 2025-01-09
animations: fix overshoot beziers and cleanup animation config parsing

PaideiaDilemma:
The change in the animationmgr is the same as https://github.com/hyprwm/Hyprland/pull/9000.

Also adapted the simplification in `CConfigManager::handleAnimation` from Hyprland.


vaxerski (approved):
thanks!

PR #644, opened 2025-01-11
animation: allow adding vars during ::tick

PaideiaDilemma:
Same as https://github.com/hyprwm/Hyprland/pull/9030


PaideiaDilemma:
Currently hyprlock does not add any vars via callbacks, but we might in the future.

vaxerski (approved):
39

PR #647, opened 2025-01-13
input-field: improve dynamic width

PaideiaDilemma:
In https://github.com/hyprwm/hyprlock/pull/582 I changed it so the placeholder text is not rendered until the dynamic width resizes and the placeholder actually fits.
That however also looks off, especially with slow animation speeds.

This PR changes it so that it will render it right away, but cut it where the input field ends.
Alba and me also thought about fading in the placeholder text, but I think that is smth for a separate PR.

CC @alba4k you wanna test again? I did not include a pad to the input field like you suggested, cause if think it looks better like that.

### DIFF
--- a/src/renderer/widgets/PasswordInputField.cpp
+++ b/src/renderer/widgets/PasswordInputField.cpp
@@ -217,21 +217,21 @@ bool CPasswordInputField::draw(const SRenderData& data) {
             }
         }
 
-        const auto CURRDOTS     = dots.currentAmount->value();
-        const int  DOTPAD       = (inputFieldBox.h - passSize.y) / 2;
-        const int  DOTAREAWIDTH = inputFieldBox.w - DOTPAD * 2;
-        const int  MAXDOTS      = std::round(DOTAREAWIDTH * 1.0 / (passSize.x + passSpacing));
-        const int  DOTFLOORED   = std::floor(CURRDOTS);
-        const auto DOTALPHA     = fontCol.a;
+        const auto   CURRDOTS     = dots.currentAmount->value();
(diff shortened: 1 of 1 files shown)


alba4k:
will test this evening

PaideiaDilemma:
> I don't really like how the shadow disappears during resizing, is that a choice you made because of performance losses?

Yep, we need to do some low-hanging shadow optimizations (like not using the entire viewport for the shadow framebuffer) before considering enabling it again I think.

> Also, is it just my eyes tricking me or did the fade animation disappear during the change of the input-field border color? pretty sure it's just changing instantly

Actually I am a bit confused about that. When i increase the animation duration, it clearly fades, but on the default config it really looks instant. However it does fade a bit still. Not really sure if it's just perception or something is slightly off...

PaideiaDilemma:
Both things are not directly related to this pr.

PaideiaDilemma:
For reference this is how looks with increased width animation duration.

https://github.com/user-attachments/assets/0f61e974-cb8a-4693-bc0e-c3811dff09f5

alba4k:
have you tried putting a video of the normal thing in slow motion?

PaideiaDilemma:
good idea.
There are a view frames of a transition in your vid or am i mistaken?
the shadow thing looks unfortunate. especially when it has to resize very slightly like in your example.

PaideiaDilemma:
@vaxerski Can we get this merged?

PaideiaDilemma:
I thought you still want to sign off on non bugfix stuff like this.

will happily merge smth like this myself in the future

Issue #572, opened 2024-12-01
Screen not updating unless Mouse is moved or Key is pressed


JoMaZu787:
### Regression?

No

### Hyprlock Info and Version

0.5.0
<details> <summary>Hyprlock config</summary>

```sh
background {
    path=/home/jomazu787/Bilder/Wallpapers/arch-btw-blender.png
    blur_size=5
    blur_passes=2
}

input-field {
    fade_on_empty=false
    outer_color=rgba(33ccffee)
    outline_thickness=2
    rounding=10
}
```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.45.2 built from branch  at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: Tue Nov 19 21:47:18 2024
Tag: v0.45.2, commits: 5451
built against aquamarine 0.5.0


flags set:
debug


System Information:
System name: Linux
Node name: desktop-jomazu787
Release: 6.11.9-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Sun, 17 Nov 2024 16:06:17 +0000


GPU information:
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA102 [GeForce RTX 3080] [10de:2206] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  565.57.01  Thu Oct 10 12:29:05 UTC 2024


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:
```

</details>

### Description

When I lock the screen with hyprlock, the display only updates either while the mouse is moving, or for a very short moment whenever a key is pressed.

### How to reproduce

1. run hyprlock
2. type text into the input field
3. the typing animation does not fully play unless the mouse is moving

### Crash reports, logs, images, videos

[...]

Issue #525, opened 2024-10-23
hyprlock not taking focus when waking up from suspend


tms-meister:
### Regression?

No

### Hyprlock Info and Version

v0.4.1-1 (AUR)
<details> <summary>Hyprlock config</summary>

```sh
$text_color = rgba(E2E2E2FF)
$entry_background_color = rgba(13131311)
$entry_border_color = rgba(91919155)
$entry_color = rgba(C6C6C6FF)
$font_family = Gabarito
$font_family_clock = Gabarito
$font_material_symbols = Material Symbols Rounded

background {
	color = rgba(20202050)
}

input-field {
	monitor = 
	size = 250, 50
	dots_size = 0.1
	dots_spacing = 0.3
	position = 0, 20
	halign = center
	valign = center
}

label { # Clock
	monitor = 
	text = $TIME
	font_size = 65
	position = 0, 300
	halign = center
	valign = center
}

label { # Greeting
    monitor =
    text = hi $USER
    shadow_passes = 1
    shadow_boost = 0.5
    color = $text_color
    font_size = 20
    font_family = $font_family

    position = 0, 240
    halign = center
    valign = center
}

label { # lock icon
    monitor =
    text = lock
    shadow_passes = 1
    shadow_boost = 0.5
    color = $text_color
    font_size = 21
    font_family = $font_material_symbols

    position = 0, 65
    halign = center
    valign = bottom
}

label { # "locked" text
    monitor =
    text = locked
    shadow_passes = 1
    shadow_boost = 0.5
    color = $text_color
    font_size = 14
    font_family = $font_family

    position = 0, 50
    halign = center
    valign = bottom
}

```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland, built from branch  at commit 4520b30d498daca8079365bdb909a8dea38e8d55  (version: bump to 0.44.1).
Date: Wed Oct 9 12:54:39 2024
Tag: v0.44.1, commits: 5309
built against aquamarine 0.4.2


flags: (if any)


System Information:
System name: Linux
Node name: puspusita
Release: 6.11.4-zen2-1-zen
Version: #1 ZEN SMP PREEMPT_DYNAMIC Sun, 20 Oct 2024 18:20:07 +0000


GPU information:
2b:00.0 VGA compatible controller [0300]: NVIDIA Corporation GA104 [GeForce RTX 3060 Ti] [10de:2486] (rev a1) (prog-if 00 [VGA controller])
[...]

Issue #419, opened 2024-07-13
Lockscreen Freeze


ghost:
Bit hard to describe but the following happens:

if i go into lock screen or my pc goes into lockscreen naturally and leave my pc for a while, i can come back and put in my password normally but the moment i press enter the pc completely freezes and i need to manually restart. i did an artifical test and the moment the clock that is on my screen freezes, is also the moment where i will have to restart manually if enter my password. Before that i can put in my password and it will resume normally

megaaa13:
Got the same issue :
Sometimes, when hyprlock is exiting, the screen just freeze, and I have to go to another tty then go back in order to get it working.
When swayidle starts hyprlock, I noticed than the clock seems to update only when I move my mouse, and was freezed at the time of locking.
Tried and sucessfully remade it in a shell, but no unusual logs. I also checked journalctl but the only logs unusual were :

> 
> Jul 15 01:49:07 tower hyprlock[4214]: pam_unix(hyprlock:auth): conversation failed
> Jul 15 01:49:07 tower hyprlock[4214]: pam_unix(hyprlock:auth): auth could not identify password for [martin]

I use Nvidia driver 555 and NixOS 24.05.
Here is my [hyprland.conf](https://github.com/megaaa13/nix-dots/blob/8e30544e5d5366b7b7840b655cb95ce9ff93bfe1/home/hyprland/hyprlock/hyprlock.conf).

megaaa13:
I am actually on 0.3.0-unstable-2024-04-24 as I'm following the Nix 24.05 channel (see (here)[https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=hyprlock]).

And the freezing could be due to a Nvidia issue, but without logs I understand it's like completely impossible to solve it...

PaideiaDilemma:
If you are able to, please try latest git (hyprland + hyprlock) and check if the freeze still happens.

I kinda think that has already been fixed.

wosym:
@PaideiaDilemma what commit made the fix? What version would contain it? I currently have 0.4.1-1 and I have more or less the same issue as described here.

PerchunPak:
For me, 0.4.1 fixed this issue, so just install 0.4.1 instead of git version

wosym:
I just installed it for the first time ever, that version, and it's affected.

PaideiaDilemma:
This is related to your Hyprland version, not hyprlock.

megaaa13:
[...]

Issue #128, opened 2024-03-03
[nvidia] Intermittently freezing for few seconds after lock


brettalcox:
Building at commit 19d7844f040c0290e23efc2e756ff75ce844248e

GPU: NVIDIA GeForce RTX 4090
WM: Hyprland
Resolution: 2560x1440 (1x scale), 3840x2160 (1.5x scale)

Added the logs below of when this happens. The screen is still a bit transparent so i can still things occurring, and the last log before the freeze is the `[ERR] Invalid key down event (stray release event?)`. The screen will go slightly transparent, and freeze for 3-5 seconds before finally finishing the lock. Other times, it doesn't happen and its a nice smooth transition/fade out. 

Weirdly, I can't get this to actually reproduce when running `hyprlock --verbose`, only simply when running `hyprlock`, so not sure if this is a race condition. 

```
gentoo :: ~ » hyprlock
[LOG]   | got iface: wl_shm v1
[LOG]   | got iface: wl_drm v2
[LOG]   | got iface: zwp_linux_dmabuf_v1 v4
[LOG]    > Bound to zwp_linux_dmabuf_v1 v4
[LOG]   | got iface: wl_compositor v6
[LOG]    > Bound to wl_compositor v6
[LOG]   | got iface: wl_subcompositor v1
[LOG]   | got iface: wl_data_device_manager v3
[LOG]   | got iface: zwlr_export_dmabuf_manager_v1 v1
[LOG]   | got iface: zwlr_data_control_manager_v1 v2
[LOG]   | got iface: zwp_primary_selection_device_manager_v1 v1
[LOG]   | got iface: wp_viewporter v1
[LOG]    > Bound to wp_viewporter v1
[LOG]   | got iface: zwlr_gamma_control_manager_v1 v1
[LOG]   | got iface: zwlr_output_power_manager_v1 v1
[LOG]   | got iface: xdg_wm_base v6
[LOG]   | got iface: wl_seat v9
[LOG]    > Bound to wl_seat v9
[LOG]   | got iface: wp_presentation v1
[LOG]   | got iface: ext_idle_notifier_v1 v1
[LOG]   | got iface: zwlr_layer_shell_v1 v4
[LOG]   | got iface: org_kde_kwin_server_decoration_manager v1
[LOG]   | got iface: zxdg_decoration_manager_v1 v1
[LOG]   | got iface: zwlr_output_manager_v1 v4
[...]

PR #655, opened 2025-01-21
core: move wayland event reading into the poll thread

PaideiaDilemma:
This was done, so that we can

  wl_display_prepare_read -> poll -> wl_display_read_events

That fixes synchronization issues on nvidia proprietary drivers.

Closes https://github.com/hyprwm/hyprlock/issues/572
Closes https://github.com/hyprwm/hyprlock/issues/525
Closes https://github.com/hyprwm/hyprlock/issues/419
Closes https://github.com/hyprwm/hyprlock/issues/128

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -117,8 +117,6 @@ void CSessionLockSurface::onScaleUpdate() {
 }
 
 void CSessionLockSurface::render() {
-    Debug::log(TRACE, "render lock");
-
     if (frameCallback || !readyForFrame) {
         needsFrame = true;
         return;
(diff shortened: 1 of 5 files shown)


PaideiaDilemma:
(mentioned the wrong issue, this does not need a wiki pr)

PaideiaDilemma:
I tested with a 3060ti using 550.142 drivers and with a very old 3rd gen intel laptop both on Hyprland and sway

vaxerski (approved):
if it works, it works.

Worth noting almost all hypr* apps use the same poll loop, so maybe it would be worthwhile to fix this in others too? (e.g. xdph)

Issue #59, opened 2024-02-22
[nvidia] Screenshot feature seems to be broken


Poly-cryptic:
The screenshot feature seems to broken for me and instead of a screenshot, a gray background pops up. These errors are probably related.

> [LOG] [gl] GL_INVALID_OPERATION error generated. <image> and <target> are incompatible
> [LOG] [gl] GL_INVALID_OPERATION error generated. <image> and <target> are incompatible
> [LOG] [gl] GL_INVALID_OPERATION error generated. <image> and <target> are incompatible

Here is my config:

> background {
>     monitor =
>     path = screenshot
> }

I have attached a full verbose log.

[log.txt](https://github.com/hyprwm/hyprlock/files/14366652/log.txt)

vaxerski:
nvidia?

Poly-cryptic:
Yeah, NVidia, I don't know how I forgot that crucial detail.

vaxerski:
https://github.com/ValveSoftware/gamescope/issues/708#issuecomment-1436051840

I am unsure how to check if a modifier is marked as such, I've asked on IRC. Will try to figure something out.

vaxerski:
[patch4.txt](https://github.com/hyprwm/hyprlock/files/14367145/patch4.txt)
hey hey try this

Poly-cryptic:
It doesn't seem to change anything whatsoever.

[log-patch1.txt](https://github.com/hyprwm/hyprlock/files/14367212/log-patch1.txt)

vaxerski:
:(

vaxerski:
[patch69.txt](https://github.com/hyprwm/hyprlock/files/14367344/patch69.txt)
try this :D

Poly-cryptic:
Nothing changed again sadly...

[log-patch2.txt](https://github.com/hyprwm/hyprlock/files/14367416/log-patch2.txt)

vaxerski:
[patch69.txt](https://github.com/hyprwm/hyprlock/files/14367749/patch69.txt)
wont fix anything but add a few logs

Poly-cryptic:
[log-patch3.txt](https://github.com/hyprwm/hyprlock/files/14367808/log-patch3.txt)

vaxerski:
are you on 10bit?

Poly-cryptic:
To be honest, I don't know. How should I go about checking that?

vaxerski:
[...]

PR #656, opened 2025-01-21
core: fix background screenshot on nvidia

PaideiaDilemma:
Fixes screencopy on nvidia by falling back to a shm buffer when dma failes. 

Closes https://github.com/hyprwm/hyprlock/issues/59


PaideiaDilemma:
I guess we probably want a option to enable it per default, so that we don't have to wait for dma to fail.

PaideiaDilemma:
I tested with a 3060ti and 550.142 drivers on Hyprland (10-bit as well) and sway (sends a different shm format, tested 24->32).

.gdb_history
```diff
@@ -0,0 +1,2 @@
+bt
+bt
```
vaxerski, line 2: lmao nice one

src/renderer/Screencopy.cpp
```diff
@@ -0,0 +1,482 @@
...
+            close(m_fd[plane]);
+
+        return false;
+    }
+
+    return true;
+}
+
+bool CSCDMAFrame::onBufferReady(SPreloadedAsset& asset) {
+    static const int general_attribs    = 3;
```
vaxerski, line 227: what is this snake doing in my repo

src/renderer/Screencopy.cpp
```diff
@@ -0,0 +1,482 @@
...
+            onFailed();
+            return;
+        }
+
+        m_sc.reset();
+    });
+}
+
+void CScreencopyFrame::onFailed() {
+    static auto* const PSCMODE = (Hyprlang::INT* const*)g_pConfigManager->getValuePtr("general:screencopy_mode");
```
vaxerski, line 82: didnt we move to the new C++-style vars from hyprlang in hyprlock? If not, might be something for another MR.

vaxerski (commented):
generally dma on nvidia should work and there's likely a mistake in the code somewhere. I mean, aquamarine creates buffers just fine.

I guess this is a fair enough compromise for now though (shm is slooow)

PaideiaDilemma:
you mean it should work with non-linear modifiers??

src/renderer/Screencopy.cpp
PaideiaDilemma: wasn't me

vaxerski:
yeah, nvidia can render to non-linear.

PaideiaDilemma:
... but it can't bind them to GL_TEXTURE_2D right?
And when using GL_TEXTURE_EXTERNAL_OES, i do get a frame for non linear mods and can render it, but there are graphical glitches, that i assumed to be cause egl cannot handle non linear. Did i get smth wrong here?

vaxerski:
no, external is needed for linear. Most modifiers should be fine with GL_TEXTURE_2D. See
https://github.com/hyprwm/Hyprland/blob/d335c8f10135e165a1085fd14134c659e4caf05a/src/render/OpenGL.cpp#L379-L422

PaideiaDilemma:
Using `eglCreateDRMImageMESA` instead of `eglCreateImage` works. Probably because the buffer get's exported via mesa and not nvidia.

PaideiaDilemma:
Ok since dma also works for nvidia now, I removed the `auto` option from `general:screencopy_mode`.
We will keep the cpu buffer implementation just in case someone needs it.

PaideiaDilemma:
Ok so on my old intel laptop `eglCreateDRMImageMESA` did not work. presumably because the frame was not exported by mesa but by an intel driver??

It now tries `eglCreateImage` if the mesa one failed to fix that.

PaideiaDilemma:
So in the end we were just missing  `EGL_DMA_BUF_PLANE0_MODIFIER_LO_EXT` and `EGL_DMA_BUF_PLANE0_MODIFIER_HI_EXT`. No need for `eglCreateDRMImageMesa`.
Thankfully vaxry pointed me to the aquamarine source, where I noticed that was missing.

PR #673, opened 2025-02-02
input-field: Fix crash when numlock is on and numlock_color is fallback

nasirHo:
## Issue
Hyprlock crashes in `updateColors()` when:
- Numlock is enabled.
- `colorConfig.num` is a fallback value.

There is an inconsistency in condition checks when assigning and later using `targetGrad`:

1. **When assigning `targetGrad`** (Line 409), the condition checks both:
   - Whether numlock is enabled.
   - Whether `colorConfig.num` is **not** a fallback.

2. **When using `targetGrad`** (Line 421), the condition only checks for numlock, but **does not** check if `numlock_color` is a fallback.  
   - If `numlock_color` **is** a fallback, `targetGrad` remains `nullptr`, leading to a crash.

## Fix
- Updated the condition in Line 421 to **match** the condition used when assigning `targetGrad`.


PaideiaDilemma (approved):

PaideiaDilemma:
Thanks!

Issue #684, opened 2025-02-08
Unplugging/Removing an output may result in dangling pointers


PaideiaDilemma:
### Regression?

No

### Hyprlock Info and Version

Hyprlock 0.6.0-git (dc6d72158c22d802845679e783231061b717ffd1)

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.47.0 built from branch  at commit 54441e0c4e51dd182f78876c014446d5d0359ba8  ().
Date: 2025-02-06
Tag: v0.47.0, commits: 
built against:
 aquamarine 0.7.2
 hyprlang 0.6.0
 hyprutils 0.5.0
 hyprcursor 0.1.11
 hyprgraphics 0.1.1


no flags were set


System Information:
System name: Linux
Node name: desktop
Release: 6.13.0
Version: #1-NixOS SMP PREEMPT_DYNAMIC Sun Jan 19 23:51:45 UTC 2025


GPU information: 
03:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6650 XT / 6700S / 6800S] [1002:73ef] (rev c1) (prog-if 00 [VGA controller])
04:00.0 Non-Volatile memory controller [0108]: Sandisk Corp SanDisk Ultra 3D / WD Blue SN570 NVMe SSD (DRAM-less) [15b7:501a] (prog-if 02 [NVM Express])
	Subsystem: Sandisk Corp SanDisk Ultra 3D / WD Blue SN570 NVMe SSD (DRAM-less) [15b7:501a]
0e:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] Raphael [1002:164e] (rev cb) (prog-if 00 [VGA controller])


os-release: ANSI_COLOR="0;38;2;126;186;228"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="25.05.20250204.799ba5b"
CPE_NAME="cpe:/o:nixos:nixos:25.05"
DEFAULT_HOSTNAME=nixos
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
ID_LIKE=""
IMAGE_ID=""
IMAGE_VERSION=""
LOGO="nix-snowflake"
NAME=NixOS
[...]

PR #686, opened 2025-02-12
core: more hyprutils smart pointer usage and safe references to widgets

PaideiaDilemma:
Closes https://github.com/hyprwm/hyprlock/issues/684

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -11,10 +11,10 @@
 CAuth::CAuth() {
     static const auto ENABLEPAM = g_pConfigManager->getValue<Hyprlang::INT>("auth:pam:enabled");
     if (*ENABLEPAM)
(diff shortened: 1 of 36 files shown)


PaideiaDilemma:
I couldn't hold myself and renamed a bunch of stuff making this way to big.
Will undo the renaming and then move it out of draft.

Maybe we should get that before 0.7.0.

Stuff I am not that happy with yet:
- `registerSelf` can't be virtual, but we need `m_self` to safely create a ref to the current widget.
- using lambdas for the callbacks kind of makes the static handler methods in the widgets redundant. Perhaps removing them and just calling the widgets methods in the lambda directly is better.

PaideiaDilemma:
@vaxerski I think this is ready

vaxerski (approved):
at your discretion

Issue #651, opened 2025-01-17
Input field color transitions look as if they are instant


PaideiaDilemma:
### Regression?

Yes

### Hyprlock Info and Version

02639c27599b03c81dc1d0e031628224f5ad8d50

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
<Paste the output of the command here>
```

</details>

### Description

Concerns the default animation config.

When i increase the duration it clearly fades between gradients.
Not really sure if it's just perception or something is slightly off...

First discussed in https://github.com/hyprwm/hyprlock/pull/647

### How to reproduce

Launch hyprlock on latest commit without explicit animation changes.

PR #689, opened 2025-02-19
animations: linear bezier for gradient, warp behavior

PaideiaDilemma:
Closes https://github.com/hyprwm/hyprlock/issues/651


PaideiaDilemma:
And with this, I think it's a good point to release 0.7.0.

PR #714, opened 2025-03-20
renderer: remove loading bar

davc0n:


PaideiaDilemma (approved):

vaxerski:
y?

PaideiaDilemma:
> y?

No visible effect on most configurations and systems.
That's why I would be in favor of removing it entirely.

But we can also keep it. In that case I think it needs some improvements.
I would expect as soon as users would actually notice it, they would either want to disable it or have it be configurable, at least in terms of color.

vaxerski (approved):
ok then

PR #718, opened 2025-03-25
core: move fail_timeout from input-field to general

PaideiaDilemma:
Initial version of this PR intended to fix some edge cases when rapidly providing input, where you would see the fail color, but no fail text placeholder.

Now it does that plus moving the `input-field:fail_timeout` setting.
It is easier and more flexible to handle the duration for which an auth failure gets displayed outside of the PasswordInputField widget.
Now per input-field fail_timeout settings wont be supported, but that didn't work anyways (smallest `fail_timeout` always took precedence)

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -76,6 +76,14 @@ void CAuth::terminate() {
     }
 }
 
+static void passwordUnlockCallback(std::shared_ptr<CTimer> self, void* data) {
+    g_pHyprlock->unlock();
+}
+
+void CAuth::enqueueUnlock() {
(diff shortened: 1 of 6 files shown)


PaideiaDilemma:
So `input-field:fail_timeout` -> `auth:fail_timeout`.
Or it should be `general:fail_timeout`??

vaxerski:
up to you

PaideiaDilemma:
I feel a bit bad about the breaking change, but I think it's just necessary.
It sucks having that within the input-field.

@vaxerski for merge cause related wiki mr

vaxerski (approved):

Issue #695, opened 2025-02-23
On wake up after suspend, hyprlock has no UI and cannot be unlocked.


ogisan-7:
### Regression?

No

### Hyprlock Info and Version

0.7.0
<details> <summary>Hyprlock config</summary>

```sh
general {
    grace = 0
    hide_cursor =  true
    no_fade_in = true
    no_fade_out = true
}

auth {
    pam:enabled = true
}

background {
    monitor = DP-1
    color = rgba(25, 23, 36, 1)  # Base color (dark background)
}

input-field {
    monitor = DP-1
    size = 20%, 5%
    outline_thickness = 1
    inner_color = rgba(25, 23, 36, 0.0)  # Transparent inner color
    placeholder_text = type here

    outer_color = rgba(235, 188, 186, 1)  # Rose color for outer border
    check_color = rgba(156, 207, 216, 1)  # Pine color for check
    fail_color = rgba(235, 111, 146, 1)   # Fail color (reddish)
    fail_text = password incorrect

    dots_rounding = 0
    dots_text_format = *
    dots_fade_time = 0
    dots_center = false

    font_color = rgb(224, 222, 244)  # Text color (light)
    fade_on_empty = false
    rounding = 0

    position = 0, -20
    halign = center
    valign = center
    font_family = Iosevka Term
}

label {
    monitor = DP-1
    text = cmd[] uname --kernel-release
    color = rgba(224, 222, 244, 1.0)  # Text color (light)
    font_size = 16
    font_family = Iosevka Term 

    position = 0, -150
    halign = center
    valign = center
}

label {
    monitor = DP-1
    text = cmd[update:1000] date +"%H:%M"
    color = rgba(156, 207, 216, 1.0)  # Pine color for time
    font_size = 55
    font_family = Iosevka Term 
    position = 0, 250
    halign = center
    valign = center
}

label {
    monitor = DP-1
    text = $USER
    color = rgba(235, 188, 186, 1.0)  # Rose color for username
    font_size = 25
    font_family = Iosevka Term

    position = 0, 80
    halign = center
    valign = center
}

shape {
    monitor = DP-1
    size = 40%, 35%
    position = 0, 0
    color = rgba(25, 23, 36, 0)  # Transparent background
    rounding = 0
    border_size = 1
    border_color = rgba(224, 222, 244, 0.8)  # Light border color
}

```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
[...]

PR #720, opened 2025-03-26
core: remove attemptRestoreOnDeath

PaideiaDilemma:
This causes confusion when trying to help people debug or get traces.
I think now that we have the lock dead screen, where recovery options are stated, we don't need this any more.

It can also cause some leakage of the desktop or other weird behavior combined with other bugs. See #695

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -39,10 +39,7 @@ CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const b
     setMallocThreshold();
 
     m_sWaylandState.display = wl_display_connect(wlDisplay.empty() ? nullptr : wlDisplay.c_str());
-    if (!m_sWaylandState.display) {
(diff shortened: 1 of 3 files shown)


src/core/hyprlock.cpp
```diff
@@ -244,10 +227,8 @@ void CHyprlock::addDmabufListener() {
         memcpy(&device, device_arr->data, sizeof(device));
 
         drmDevice* drmDev;
-        if (drmGetDeviceFromDevId(device, /* flags */ 0, &drmDev) != 0) {
-            Debug::log(WARN, "[dmabuf] unable to open main device?");
-            exit(1);
-        }
+        if (drmGetDeviceFromDevId(device, /* flags */ 0, &drmDev) != 0)
+            RASSERT(false, "[dmabuf] unable to open main device?");
```
vaxerski, line 231: the if expr can be inside rassert

vaxerski (approved):

Issue #702, opened 2025-03-06
FIngerprint auth does not work if session is left locked for a long time


KartikayD25:
### Regression?

No

### Hyprlock Info and Version

<Hyprlock version v0.7.0>
<details> <summary>Hyprlock config</summary>

```sh
# BACKGROUND
background {
    monitor =
    path = $HOME/Pictures/Wallpaper-Bank/wallpapers/background
    blur_passes = 2
    contrast = 0.8916
    brightness = 0.8172
    vibrancy = 0.1696
    vibrancy_darkness = 0.0
}

# GENERAL
general {
    no_fade_in = false
    grace = 0
    disable_loading_bar = false
}

auth {
    fingerprint{
        enabled = true
      }
  }

# INPUT FIELD
input-field {
    monitor =
    size = 250, 60
    outline_thickness = 2
    dots_size = 0.2 # Scale of input-field height, 0.2 - 0.8
    dots_spacing = 0.2 # Scale of dots' absolute size, 0.0 - 1.0
    dots_center = true
    outer_color = rgba(0, 0, 0, 0)
    inner_color = rgba(100, 114, 125, 0.4)
    font_color = rgb(200, 200, 200)
    fade_on_empty = false
    font_family = SF Pro Display Bold
    placeholder_text = <i><span foreground="##ffffff99">Enter Pass</span></i>
    hide_input = false
    position = 0, -225
    halign = center
    valign = center
}

# Time
label {
    monitor =
    text = cmd[update:1000] echo "<span>$(date +"%H:%M")</span>"
    color = rgba(216, 222, 233, 0.70)
    font_size = 130
    font_family = SF Pro Display Bold
    position = 0, 240
    halign = center
    valign = center
}

# Day-Month-Date
label {
    monitor =
    text = cmd[update:1000] echo -e "$(date +"%A, %d %B")"
    color = rgba(216, 222, 233, 0.70)
    font_size = 30
    font_family = SF Pro Display Bold
    position = 0, 105
    halign = center
    valign = center
}

# Profie-Photo
image {
    monitor =
    path = ~/Pictures/profile.jpg
    border_color = 0xffdddddd
    border_size = 0
    size = 120
    rounding = -1
    rotate = 0
    reload_time = -1
    reload_cmd = 
    position = 0, -20
    halign = center
    valign = center
}

# USER
label {
    monitor =
    text = Hi, $USER
    color = rgba(216, 222, 233, 0.70)
    font_size = 25
[...]

PR #722, opened 2025-03-27
fingerprint: allow fprint to suspend and cancel verify for us

BlueZeeKing:
fprint does have it's own logic to inhibit sleeping until it cancels active tasks. I believe that hyprlock's logic to stop the verification before sleeping was causing an issue within fprint as it tried to handle both at the same time.

I was only having this issue occasionally so I'm not sure this completely fixes it, which is why this is a draft.

### DIFF
--- a/src/auth/Fingerprint.cpp
+++ b/src/auth/Fingerprint.cpp
@@ -65,18 +65,12 @@ void CFingerprint::init() {
         // When entering sleep, the wake signal will trigger startVerify().
         if (m_sDBUSState.sleeping)
             return;
-        inhibitSleep();
         startVerify();
     });
     m_sDBUSState.login->uponSignal("PrepareForSleep").onInterface(LOGIN_MANAGER).call([this](bool start) {
         Debug::log(LOG, "fprint: PrepareForSleep (start: {})", start);
-        if (start) {
-            m_sDBUSState.sleeping = true;
-            stopVerify();
-            m_sDBUSState.inhibitLock.reset();
-        } else {
-            m_sDBUSState.sleeping = false;
-            inhibitSleep();
+        m_sDBUSState.sleeping = start;
+        if (!m_sDBUSState.sleeping && !m_sDBUSState.verifying) {
             startVerify();
         }
     });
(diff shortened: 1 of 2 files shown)


BlueZeeKing:
This should fix #702

src/auth/Fingerprint.cpp
```diff
@@ -65,18 +65,12 @@ void CFingerprint::init() {
...
-            m_sDBUSState.sleeping = true;
-            stopVerify();
-            m_sDBUSState.inhibitLock.reset();
-        } else {
-            m_sDBUSState.sleeping = false;
-            inhibitSleep();
+        m_sDBUSState.sleeping = start;
+        if (!m_sDBUSState.sleeping && !m_sDBUSState.verifying) {
             startVerify();
         }
```
PaideiaDilemma, line 75: I mean this might fix, but who knows what fingerprint drivers do when the system suspends. So maybe we should stop verify and release the device too? Perhaps that would be a safer fix?

src/auth/Fingerprint.cpp
```diff
@@ -65,18 +65,12 @@ void CFingerprint::init() {
...
         Debug::log(LOG, "fprint: PrepareForSleep (start: {})", start);
-        if (start) {
-            m_sDBUSState.sleeping = true;
-            stopVerify();
-            m_sDBUSState.inhibitLock.reset();
-        } else {
-            m_sDBUSState.sleeping = false;
-            inhibitSleep();
+        m_sDBUSState.sleeping = start;
+        if (!m_sDBUSState.sleeping && !m_sDBUSState.verifying) {
```
PaideiaDilemma, line 73: Braces can be removed here.

src/auth/Fingerprint.cpp
BlueZeeKing: Fprintd does a very similar thing that hyprlock does when suspending. It inhibits sleeping until all devices have suspended ([initial call to suspend](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/src/manager.c?ref_type=heads#L348), [release sleep inhibitor](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/src/manager.c?ref_type=heads#L269)). libfprint then calls the driver to suspend if possible, which means hyprlock doesn't have to do anything with the fingerprint sensor during suspend, or libfprint cancels the verify operation, so hyprlock must restart it when waking ([libfprint suspend start](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/src/manager.c?ref_type=heads#L269), [libfprint suspend complete](https://gitlab.freedesktop.org/libfprint/fprintd/-/blob/master/src/manager.c?ref_type=heads#L269) (this is the part that cancels the verify task if the driver/sensor cannot suspend itself)). Trying to stop the verification while all this is happening is what I think is causing issues, but either way it should be perfectly fine to allow fprint to cancel the operation instead of doing it manually.

PaideiaDilemma:
In case nobody from the issue checks back here, just ping me and I think we can merge it.
Worst case, we revert it back.

steveulfelder:
https://github.com/hyprwm/hyprlock/issues/702#issuecomment-2780171332

PaideiaDilemma:
Merging, since it's probably the right thing to do. Keeping the issue open for now.

PR #726, opened 2025-03-31
core: use a screenshot for fade in/out

PaideiaDilemma:
Instead of relying on Hyprland drawing it's workspaces when locked, use a screenshot to fade in and out.

Basically had to rewrite the background widget.
No merging yet. I might revision the background widget.

Needs testing.

Note:
This is supposed to go together with a MR that makes it so Hyprland does not draw workspaces when locked per default. But I have a `misc:allow_session_lock_xray` setting planed to keep it possible to use `shape:xray` or implement dynamic blur.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -6,6 +6,7 @@
 #include "../auth/Auth.hpp"
 #include "../auth/Fingerprint.hpp"
 #include "Egl.hpp"
+#include <chrono>
 #include <hyprutils/memory/UniquePtr.hpp>
 #include <sys/wait.h>
 #include <sys/poll.h>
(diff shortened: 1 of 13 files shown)


fxzzi:
just tested for ya.

This helps a lot with the fade in animation. Previously it wouldn't work on my main monitor (maybe it did, but the fade looked like it was instant)

with this change the transition is a lot smoother. Thanks!

PaideiaDilemma:
I also tested the `reload_cmd` and such. I think this is ready.

One thing I didn't mention yet: With this, the crossfade of the background can be configured with the `fadeIn` animation config, because it uses the same code as the fade in from screenshot the the inital background.
Should probably be it's own animation config in the future.

src/core/hyprlock.cpp
```diff
@@ -313,17 +314,30 @@ void CHyprlock::run() {
...
-            } else {
-                wl_display_dispatch(m_sWaylandState.display);
+    {
+        // Gather background resources and screencopy frames before locking the screen.
+        // We need to do this because as soon as we lock the screen, workspaces frames can no longer be captured. It either won't work at all, or we will capture hyprlock itself.
+        // Bypass with --immediate-render (can cause the background first rendering a solid clor and missing or inaccurate screencopy frames)
+        const auto MAXDELAYMS    = 2000; // 2 Seconds
+        const auto STARTGATHERTP = std::chrono::system_clock::now();
+        if (!g_pHyprlock->m_bImmediateRender) {
+            while (!g_pRenderer->asyncResourceGatherer->gathered) {
```
vaxerski, line 324: slow down maaan no need to peg a core. You should add that 1ms sleep back.

This is a hack in general, ideally we'd do a poll() but yeah this is easier

src/renderer/Framebuffer.cpp
```diff
@@ -113,6 +113,13 @@ void CFramebuffer::release() {
     m_pStencilTex   = nullptr;
 }
 
+void CFramebuffer::abandon() {
```
vaxerski, line 116: this should be called "leak" and at that point you start asking: why is this here? Can't a shared_ptr fix it?

b0ch3nski (commented):
Just some nitpicks

src/renderer/AsyncResourceGatherer.cpp
```diff
@@ -1,62 +1,59 @@
...
-    const auto               CWIDGETS = g_pConfigManager->getWidgetConfigs();
-
-    std::vector<std::string> mons;
-
-    for (auto& c : CWIDGETS) {
-        if (c.type != "background")
-            continue;
+    const auto FADEINCFG    = g_pConfigManager->m_AnimationTree.getConfig("fadeIn");
+    const auto FADEOUTCFG   = g_pConfigManager->m_AnimationTree.getConfig("fadeOut");
+    const auto BGSCREENSHOT = std::ranges::any_of(g_pConfigManager->getWidgetConfigs(), [](const auto& w) { //
```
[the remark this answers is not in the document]
PaideiaDilemma: nah that's for formating...

vaxerski:
because it's not valid regex probably. The first parameter is a _regex_, not a path.

leon-erd:
I have set this in nix as per the wiki
```nix
    permission = [
      "${lib.getExe config.programs.hyprlock.package}, screencopy, allow"
    ];
```
This just expands to the line I posted.
Doing the same for grim etc works no problem. Could it be that internally hyprlock is spawning a different process to capture the screen which is not allowed by the rule?

Anyways I will try the regex from the wiki just to be sure

leon-erd:
Ahhh I see what the problem is!! Characters with special meaning for regex (e.g. `+`) are not escaped and therefore the regex doesn't match the string. Makes total sense now :grinning: 
Nix has a function for this. I will submit a PR for the wiki making use of this :))

fufexan:
Damn, forgot we have + in some version strings.

PR #730, opened 2025-04-07
fingerprint: update widgets after changing prompt

BlueZeeKing:
When verification fails and a retry is needed (like when the user didn't hold
their finger for long enough) the prompt variable is changed but the widgets
aren't actually updated so the new prompt is never shown to the user. This fixes that.


PaideiaDilemma:
makes sense

PaideiaDilemma (approved):

Issue #366, opened 2024-06-13
Feature Request: On-Click  Command Support in Hyprlock Widget


ashish-kus:
I hope this message finds you well. I am writing to request a new feature for Hyprlock that would significantly enhance its functionality and user experience.

## Feature Request:
I would like to propose the addition of an on-click bash command feature within the widget in Hyprlock. This functionality would allow users to execute bash commands directly from the lockscreen, enabling actions such as **shutdown, reboot, logout, and sleep** with a single click.

#### Suggested Implementation:
- Add an option in the widget configuration to specify bash commands for different on-click actions.
- Ensure secure execution of these commands to prevent any potential misuse
    
    
Example configuration snippet:
``` 
widget_name {
monitor =
# further options
on-click = systemctl syspend 
}
```

Eclextic:
Well, what is secure execution? I’m totally on board, but what do you propose as a security measure?

Noino:
Isn't a button on the lock screen by definition unauthorized, so whatever the user here exposes is on them imho. 

Totally want a suspend button though that i can press even though I'm carrying a child or two

Eclextic:
I think maybe it's better to have a list of commands that can be executed and only those?

Mainly for security as well as testability with bugs, minimizing them etc...

And we could have it so that there is another variable that allows bash script execution, but whenever that is used, we get a warning and need to supply a flag or smth...


So what I propose:
2 Variables. 1 uses a list of known simple bash commands that are known to be safe and can't be executed without notifying the user.
2 allows for bash script execution, but using them forces a flag to make sure that the user is responsible for what is about to happen next. (Then again bash scripts already get executed with no verification sooo... Make of it what you will.)

bvr-yr:
security what? it's totally responsibility of users and their own config, if they wish it could be even `rm -rf --no-preserve-root /`, which you even now technically can execute in label `cmd` for example. By simply clicking button, you can't do anything beyond what is already defined and therefor permitted by the user in config

SERDUSSS:
This feature would open an infinite number of possibilities and I think these should be kept without any limitation

Also this should not suppose a unauthorized command execution since as @Noino said, if you let someone execute a command without being authenticated you should be sure that command will not cause a threat to the system and that it cannot be abused in any way

bvr-yr:
please explain me how can one abuse anthing that is permitted by user since this command is predefined in config, I can't get it

ejiektpobehuk:
[...]

PR #736, opened 2025-04-10
widgets: add onclick feature

PaideiaDilemma:
Picks up the contribution from @Memoraike (thanks :))
Implements https://github.com/hyprwm/hyprlock/issues/366.

### DIFF
--- a/src/core/CursorShape.cpp
+++ b/src/core/CursorShape.cpp
@@ -9,8 +9,13 @@ CCursorShape::CCursorShape(SP<CCWpCursorShapeManagerV1> mgr) : mgr(mgr) {
 }
 
 void CCursorShape::setShape(const wpCursorShapeDeviceV1Shape shape) {
-    if (!dev)
+    if (!g_pSeatManager->m_pPointer) {
         return;
+    }
+
+    if (!dev) {
+        dev = makeShared<CCWpCursorShapeDeviceV1>(mgr->sendGetPointer(g_pSeatManager->m_pPointer->resource()));
+    }
 
     dev->sendSetShape(lastCursorSerial, shape);
 }
--- a/src/core/CursorShape.cpp
+++ b/src/core/CursorShape.cpp
@@ -9,13 +9,11 @@ CCursorShape::CCursorShape(SP<CCWpCursorShapeManagerV1> mgr) : mgr(mgr) {
 }
 
 void CCursorShape::setShape(const wpCursorShapeDeviceV1Shape shape) {
-    if (!g_pSeatManager->m_pPointer) {
+    if (!g_pSeatManager->m_pPointer)
         return;
-    }
 
-    if (!dev) {
+    if (!dev)
         dev = makeShared<CCWpCursorShapeDeviceV1>(mgr->sendGetPointer(g_pSeatManager->m_pPointer->resource()));
-    }
 
     dev->sendSetShape(lastCursorSerial, shape);
 }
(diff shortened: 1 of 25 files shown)


PaideiaDilemma:
Currently `on-click` is implemented for labels and onHover changes the cursor when on the input-field.

Do we want to implement the following things here already or in a future MR?
- `on-click` for shape and image.
- visual feedback for onClick (and maybe for onHover)

rotmh:
I don't have any say here... :)

But in my opinion:

> `on-click` for shape and image.

Should probably be implemented in this MR, because it's part of the same premise as `on-click` for labels, but:

> visual feedback for onClick (and maybe for onHover)

I think should be a part of a separate MR, because it's seems like a separate feature.

PaideiaDilemma:
Added onclick for shape and image.

Then i noticed the it triggers onclick for widgets on different monitors than my cursor is on. (I have two 1080p monitors and a single shape widget would trigger the onclick command twice)
I implemented handling for the `enter` and `leave` events of `wl_pointer` to localize onclick and hover to the currently focused output and fixed up the positions a bit.

PaideiaDilemma:
ups, sorry. didn't mean to delete the branch

JunaidQrysh:
@PaideiaDilemma can you please add visual feedback when cursor is on a label or image; i have setup reboot icon
it would be good if it says "Reboot" when the cursor is on that icon.

src/core/CursorShape.cpp
```diff
@@ -9,8 +9,13 @@ CCursorShape::CCursorShape(SP<CCWpCursorShapeManagerV1> mgr) : mgr(mgr) {
 }
 
 void CCursorShape::setShape(const wpCursorShapeDeviceV1Shape shape) {
-    if (!dev)
+    if (!g_pSeatManager->m_pPointer) {
```
vaxerski, line 12: no {} here and blow

PaideiaDilemma:
@JunaidQrysh  yeah i want that as well but not in this PR. We can add it later

src/renderer/widgets/Shape.cpp
```diff
@@ -100,3 +103,19 @@ bool CShape::draw(const SRenderData& data) {
...
+CBox CShape::getBoundingBoxWl() const {
+    return {
+        Vector2D{pos.x, viewport.y - pos.y - size.y},
+        size,
+    };
+}
+
+void CShape::onClick(uint32_t button, bool down, const Vector2D& pos) {
+    if (down && !onclickCommand.empty())
+        g_pHyprlock->spawnSync(onclickCommand);
```
vaxerski, line 115: dumb idea. someone puts `htop` by accident and hyprlock is stuck

evtn: the current [dynamic labels](https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock/#dynamic-labels) doc states that one shouldn't put never-ending commands there already, how is this different?

PaideiaDilemma: dynamic labels run in the asyncResourceGatherer.
Making that thread hang is also not ideal but not as bad as stalling the "main" thread (which handles all wayland events, etc.)

vaxerski: furthermore there is no need for sync here so it's just hanging for no reason

vaxerski (approved):
at your discretion

PR #763, opened 2025-05-05
core: disable fade in when using --immediate

### LINKED ISSUE #761: Disable fade-in when launched with `--immediate`
### Description

https://github.com/hyprwm/hyprlock/issues/633#issuecomment-2851370330

I think that makes sense.
Now that we have the `hyprland-lock-notify` protocol, we can at least guaranty that the compositor locked.
But if one uses fade-in, there is still a risk of the desktop showing for a tiny bit.

I propose to let the `--immediate` flag disable fade-in.
`--immediate` should be used for `before_sleep` commands any way (since `grace` does not make sense across a  sleep)

nnyyxxxx:
closes #761

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -107,7 +107,7 @@ int main(int argc, char** argv, char** envp) {
         return 1;
     }
 
-    if (noFadeIn)
+    if (noFadeIn || immediate)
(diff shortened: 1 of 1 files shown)


PaideiaDilemma (approved):
Thank you!

PR #781, opened 2025-05-21
core: add grace option

### LINKED ISSUE #776: Allow passing grace time using cmdline args
### Description

We have the `--immediate` option to ignore any grace time set in the config.
What if we add a `--grace` option to override the grace time defined in the config?

Example:

`hyprlock --grace 5`
Will lock with grace period set to 5 seconds.

I can make an MR if desired

HarshNarayanJha:
This adds the `--grace` option to specify a grace value that will override the one from config, when running hyprlock. We already have the `--immediate` option, which does the opposite. This can be useful when say you have grace = 0 in the config (I don't like having a grace time on lockscreen), while you want to lock with a grace just for this time.

`--grace` accepts a `long` value, which is valid if `>= 0`, with a default value of `-1`, indicating no override was passed.

I have modified the `immediate` check logic in `core/hyprlock.cpp` to address the same and load the value from config only if a valid `grace` was passed.

Closes #776

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -35,7 +35,7 @@ static void setMallocThreshold() {
 #endif
 }
 
-CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender) {
+CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const long grace) {
     setMallocThreshold();
 
     m_sWaylandState.display = wl_display_connect(wlDisplay.empty() ? nullptr : wlDisplay.c_str());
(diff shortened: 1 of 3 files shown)


PaideiaDilemma (commented):
Looks fine to me. Although, should we make it an unsigned? Currently you could provide a negative values to grace, which is fine, but also doesn't make any sense.

HarshNarayanJha:
We have a check for that, you can't pass negatives via args. I can't make it `unsigned` since I check if passed value is `-1` to skip that and load value from config. Cannot use 0 for that since `--grace 0` is a valid option to pass, and then I can no longer decide if a grace value was even passed or not, without using some extra logic.

While testing I passed some negatives, and yes it is fine, chrono accepts it happily and grace time is effectively 0!

HarshNarayanJha:
Also, since now `--immediate` and `--grace 0` do effectively the same thing, I think we should start considering this as well

> I have been thinking about this as well.
> I would even be in favor of removing grace from the config entirely, because a lot of people copy configurations with `general:grace` set without knowing what it does. Although that would be a breaking change.
> 
> Feel free to send a PR.
> And about `--immediate`: I would keep it, but now instead of influencing grace, it would be `--no-fade-in` + `--immediate-render`. That way it still expresses "I want to lock the screen ASAP".
> If we don't want that, we probably should remove `--intimidate`, which would be another breaking change.
> 
> Having `grace` specified as an argument with the default `0` seems more intuitive to me than what we have now. 

 _Originally posted by @PaideiaDilemma in [#776](https://github.com/hyprwm/hyprlock/issues/776#issuecomment-2896897717)_

PaideiaDilemma:
Makes sense. I would just remove the config option for grace in this PR.
Then we don't need to have it at -1 by default.

If you don't want to do that it's fine. I think we can merge this as is.

HarshNarayanJha:
That would be a breaking change, let's do it in a different PR.

Please merge this.

Issue #789, opened 2025-06-01
Feature: Add extra configuration option font_weight


kleinschrader:
### Description

As a hyprlock user i would like to be able to pass a new property called font_weight to adjust how bold text should be.

How i image the config would look:
```
label {
    monitor =
    text = Hi there, $USER
    color = rgba(200, 200, 200, 1.0)
    font_size = 25
    font_family = Noto Sans
    font_weight = 700

    position = 0, 80
    halign = center
    valign = center
}
```

Preferably this new property should default to `400` (Aka normal) and be able to be set on the following widgets:
- input-field
- label

Additional References:
- [Where properties are passed to pango](https://github.com/hyprwm/hyprlock/blob/da1d076d849fc0f298c1d287bddd04802bf7d0f9/src/renderer/AsyncResourceGatherer.cpp#L209)
- [Pango reference set_weight](https://docs.gtk.org/Pango/method.FontDescription.set_weight.html)
- [Label config loading](https://github.com/hyprwm/hyprlock/blob/da1d076d849fc0f298c1d287bddd04802bf7d0f9/src/renderer/widgets/Label.cpp#L68)
- [Password Input field config loading](https://github.com/hyprwm/hyprlock/blob/da1d076d849fc0f298c1d287bddd04802bf7d0f9/src/renderer/widgets/PasswordInputField.cpp#L26)
- [ConfigManager](https://github.com/hyprwm/hyprlock/blob/da1d076d849fc0f298c1d287bddd04802bf7d0f9/src/config/ConfigManager.cpp#L1)

I'll attempt to make a PR adding this myself, however it has been a while since i used C++ so the quality may be unacceptable.

PaideiaDilemma:
One can already use pango markup for that.

Let us know if you still need an extra option for whatever reason.

PR #790, opened 2025-06-01
[Feature] Add font_weight configuration option to label and password input

kleinschrader:
This adds the feature font_weight configuration option described in my issue #789 

Please scrutinize the [static cast](src/renderer/AsyncResourceGatherer.cpp#L236) as im sure there is a better way.


PaideiaDilemma:
Hmm you would already be able to adjust the font_weight by using pango markup. See [this].(https://docs.gtk.org/Pango/pango_markup.html#:~:text=normal%E2%80%99%2C%20%E2%80%98oblique%E2%80%99%2C%C2%A0%E2%80%98italic%E2%80%99.-,font_weight,-weight)

Other than the discover ability problem of pango markup, is there anything that an explicit option would alow you to do that pango markup doesn't?

kleinschrader:
Ah, i must have missed that part in the hyprlock documentation about pango markup.

I am fully ok with just closing this pr and related issue as there is no benefit to doing it this way.
Thanks for the response and taking a look!

PR #799, opened 2025-06-07
widgets: use thread safe references

PointerDilemma:
I was able to repro the crash in #785 with a script that suspends as Hyprlock starts up.

<details><summary>test script</summary>

```bash
#!/usr/bin/env bash

test() {
  sudo rtcwake -m mem -s 3 &
  suspend_pid=$!
  sleep 0.5
  ~/desk/hyprlock/build/hyprlock --config ~/.config/hypr/hyprlock/lots-of-label-updates.conf &
  hyprlock_pid=$(pidof hyprlock)

  echo "Hyprlock PID: $hyprlock_pid"

  wait "$suspend_pid"

  sleep 3

  pkill -USR1 hyprlock

  wait "$hyprlock_pid"

  # Exit if hyprlock did not exit cleanly
  if [ $? -ne 0 ]; then
    echo "Hyprlock did not exit cleanly"
    exit 1
  fi
}

while true; do
  test
done
```

</details>

When the issue happens, we have a heap corruption occurring, because some refcount is corrupted. 

<details><summary>Example crash</summary>

```bt
#0  0x000000000049788f in Hyprutils::Memory::CWeakPointer<CLabel>::decrementWeak() ()
#1  0x00000000004b2faa in std::_Function_handler<void (std::shared_ptr<CTimer>, void*), CLabel::plantTimer()::{lambda(auto:1, auto:2)#1}>::_M_manager(std::_Any_data&, std::_Any_data const&, std::_Manager_operation) ()
#2  0x0000000000475aab in std::_Sp_counted_ptr_inplace<CTimer, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
#3  0x000000000046e603 in CHyprlock::~CHyprlock() ()
#4  0x000000000047e144 in Hyprutils::Memory::Impl_::impl<CHyprlock>::destroy() ()
#5  0x000000000042bcb1 in Hyprutils::Memory::CUniquePointer<CHyprlock>::~CUniquePointer() ()
#6  0x00007f9748643bf1 in __run_exit_handlers ()
    at /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libc.so.6
#7  0x00007f9748643cb0 in exit () at /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libc.so.6
#8  0x00007f974862a485 in __libc_start_call_main ()
    at /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libc.so.6
#9  0x00007f974862a539 in __libc_start_main_impl ()
    at /nix/store/cg9s562sa33k78m63njfn1rw47dp9z0i-glibc-2.40-66/lib/libc.so.6
#10 0x0000000000429835 in _start ()
```

</details>

With https://github.com/hyprwm/hyprlock/commit/9f37c1c8e9923dff65b4b1771694be3be59b1836 I moved all references to hyprutils smart pointer, which fixed a bunch of crashes related to monitor changes.
Some references to widgets are passed to the AsyncResourceGatherer.
I thought it would be fine, since we move all the stuff to the other thread and never access or construct a new SP/WP within another thread.

However, there is a case where we have a wrong refcount when sleep happens at a specific point in time. Moving all widget pointers to STL smart pointers seems to fix the issue. (Havn't had a crash with that.)

Closes https://github.com/hyprwm/hyprlock/issues/785

Could potentially also fix some other open issues.

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -459,6 +459,8 @@ void CHyprlock::run() {
                 passed.push_back(t);
         }
 
+        timerscpy.clear();
+
         m_sLoopState.timersMutex.lock();
         std::erase_if(m_vTimers, [passed](const auto& timer) { return std::find(passed.begin(), passed.end(), timer) != passed.end(); });
         m_sLoopState.timersMutex.unlock();
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -388,13 +388,13 @@ void CRenderer::renderTextureMix(const CBox& box, const CTexture& tex, const CTe
 }
 
 template <class Widget>
-static void createWidget(std::vector<SP<IWidget>>& widgets) {
-    const auto W = makeShared<Widget>();
+static void createWidget(std::vector<std::shared_ptr<IWidget>>& widgets) {
+    const auto W = std::make_shared<Widget>();
     W->registerSelf(W);
     widgets.emplace_back(W);
 }
 
-std::vector<SP<IWidget>>& CRenderer::getOrCreateWidgetsFor(const CSessionLockSurface& surf) {
+std::vector<std::shared_ptr<IWidget>>& CRenderer::getOrCreateWidgetsFor(const CSessionLockSurface& surf) {
     RASSERT(surf.m_outputID != OUTPUT_INVALID, "Invalid output ID!");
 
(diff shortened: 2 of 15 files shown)


PaideiaDilemma:
I will try to do that : )

Issue #782, opened 2025-05-22
Removal of grace from the configuration


PaideiaDilemma:
I would like to remove the grace option from the hyprlock configuration entirely.
Instead I think it should exclusively be passed as a cli argument via `hyprlock --grace x`.

Why:
It's cleaner and it happens a bit too regularly that someone copies a configuration with grace in it and doesn't know why it just unlocks. 

Let me know if anyone thinks that's a bad idea.

PR #802, opened 2025-06-14
core: replace grace option with --grace cli argument

davc0n:
ref https://github.com/hyprwm/hyprlock/issues/782

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -43,10 +43,20 @@ CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const b
 
     g_pEGL = makeUnique<CEGL>(m_sWaylandState.display);
 
-    if (!immediate) {
-        m_tGraceEnds = graceSeconds ? std::chrono::system_clock::now() + std::chrono::seconds(graceSeconds) : std::chrono::system_clock::from_time_t(0);
-    } else
+    static const auto GRACE = g_pConfigManager->getValue<Hyprlang::INT>("general:grace");
+    if (*GRACE > 0) {
+        Debug::log(WARN, "\"general:grace\" config option is deprecated. It will be removed in an upcoming release. Use the \"--grace\" option instead.");
+    }
+
+    if (immediate) {
         m_tGraceEnds = std::chrono::system_clock::from_time_t(0);
+    } else if (graceSeconds > 0) {
+        m_tGraceEnds = std::chrono::system_clock::now() + std::chrono::seconds(graceSeconds);
+    } else if (*GRACE > 0) {
+        m_tGraceEnds = std::chrono::system_clock::now() + std::chrono::seconds(*GRACE);
+    } else {
+        m_tGraceEnds = std::chrono::system_clock::from_time_t(0);
+    }
 
     static const auto IMMEDIATERENDER = g_pConfigManager->getValue<Hyprlang::INT>("general:immediate_render");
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -77,10 +77,22 @@ int main(int argc, char** argv, char** envp) {
             else
                 return 1;
 
-        } else if (arg == "--immediate")
```
PaideiaDilemma, line 80: Keep the --immediate for now and set `graceSeconds` to 0 and log something like `Debug::log(WARN, "\"--immediate\" together with \"general:grace\" is deprecated. It will be removed in an upcoming release. Use the \"--grace\" option instead.");`

src/core/hyprlock.cpp
```diff
@@ -35,19 +35,15 @@ static void setMallocThreshold() {
...
+CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediateRender, const int graceSeconds) {
     setMallocThreshold();
 
     m_sWaylandState.display = wl_display_connect(wlDisplay.empty() ? nullptr : wlDisplay.c_str());
     RASSERT(m_sWaylandState.display, "Couldn't connect to a wayland compositor");
 
     g_pEGL = makeUnique<CEGL>(m_sWaylandState.display);
 
-    if (!immediate) {
-        static const auto GRACE = g_pConfigManager->getValue<Hyprlang::INT>("general:grace");
```
PaideiaDilemma, line 47: Also here: Do we want to log that it's deprecated?

src/core/hyprlock.cpp
```diff
@@ -43,10 +43,20 @@ CHyprlock::CHyprlock(const std::string& wlDisplay, const bool immediate, const b
...
+
+    if (immediate) {
         m_tGraceEnds = std::chrono::system_clock::from_time_t(0);
+    } else if (graceSeconds > 0) {
+        m_tGraceEnds = std::chrono::system_clock::now() + std::chrono::seconds(graceSeconds);
+    } else if (*GRACE > 0) {
+        m_tGraceEnds = std::chrono::system_clock::now() + std::chrono::seconds(*GRACE);
+    } else {
+        m_tGraceEnds = std::chrono::system_clock::from_time_t(0);
+    }
```
PaideiaDilemma, line 59: braces :)

src/core/hyprlock.hpp
```diff
@@ -29,7 +29,7 @@ struct SDMABUFModifier {
 
 class CHyprlock {
   public:
-    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender);
+    CHyprlock(const std::string& wlDisplay, const bool immediate, const bool immediateRender, const int gracePeriod);
```
PaideiaDilemma, line 32: What I meant with "keep --immediate" is just in main.cpp. Because if we wouldn't have handled the cli argument, hyprlock wouldn't start. I liked it better before, where you removed immediate from the constructor here.

src/core/hyprlock.cpp
davc0n: yeah just saw they're omitted for single line blocks

PaideiaDilemma (approved):
Top

PaideiaDilemma:
Ups, forgot the wiki MR.
@vaxerski https://github.com/hyprwm/hyprland-wiki/pull/1129https://github.com/hyprwm/hyprland-wiki/pull/1129 would need merging.

Issue #792, opened 2025-06-02
Configurable fingerprint retry limit


notpeelz:
### Description

It would be good if there was a way to increase (or disable) the hard-coded limit of 3 retries for fingerprint auth.

Related: #353

PR #833, opened 2025-07-27
Add configuration option for maximum number of fingerprint retries

coldiron:
Satisfies #792 

Adds a new configuration option `auth:fingerprint:max_retries` with a default of 3, replacing the existing hard-coded value of 3 fingerprint retires.

### DIFF
--- a/assets/example.conf
+++ b/assets/example.conf
@@ -22,6 +22,7 @@ general {
 #         ready_message = Scan fingerprint to unlock
 #         present_message = Scanning...
 #         retry_delay = 250 # in milliseconds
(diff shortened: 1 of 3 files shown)


PaideiaDilemma (approved):
Thanks I think this is good after looking at the code Fingerprint code again.

Wiki MR needed though :)

alba4k:
@PointerDilemma I guess this PR makes no sense if that kind of stuff is moving to hyprauth, but maybe add something like infinite fingerprint retries there? :)

would make sense, as biometrics aren't as easily brute-forced as passwords

PR #840, opened 2025-08-01
Make detection of PAM library more portable

tagattie:
A check for PAM library was added by https://github.com/hyprwm/hyprlock/pull/795. FreeBSD's base system has PAM implementation based on [OpenPAM](http://www.openpam.org/) and it does not provide `pam.pc` file. So `pkg_check_modules()` fails to detect PAM library.

This PR proposes a small change to how PAM library is detected by cmake. First, cmake tries to find it using `find_library()`. FreeBSD's `libpam` is in standard search paths so cmake can find it. If otherwise, cmake falls back to `pkg_check_modules()` so any system providing `pam.pc` can be supported.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,14 +86,25 @@ pkg_check_modules(
   pangocairo
   libdrm
   gbm
-  pam
   hyprutils>=0.8.0
   sdbus-c++>=2.0.0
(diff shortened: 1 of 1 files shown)


PaideiaDilemma:
Thanks @tagattie for the freebsd support :)

Issue #793, opened 2025-06-04
[NVIDIA] hyprlock crashing when waking up from suspend


Imads608:
### Regression?

Yes

### Hyprlock Info and Version

**Hyprlock Version**
Hyprlock version 0.8.2-1

<details> <summary>Hyprlock config</summary>

```sh
#! Hyprlock Boilerplate Configuration File

#  Hyprlock default configuration file handler
$BACKGROUND_PATH = $XDG_CACHE_HOME/hyde/wall.set.png
$HYPRLOCK_BACKGROUND = $XDG_CACHE_HOME/hyde/wallpapers/hyprlock.png
$SPLASH_CMD = hyprlock.sh --mpris || hyprctl splash
$MPRIS_TEXT = hyprlock.sh --mpris
$MPRIS_IMAGE = $XDG_CACHE_HOME/hyde/landing/mpris.png
$PROFILE_IMAGE = $XDG_CACHE_HOME/hyde/landing/profile.jpg
$GREET_TEXT =echo "Good $(date +%H | awk '{if ($1 < 12) print "Morning  "; else if ($1 < 18) print "Afternoon  "; else print "Evening  "}')"
$CAVA_CMD= { [ -f /tmp/hyprlock-cava ] && tail -n 1 /tmp/hyprlock-cava 2>/dev/null ;}  ||  { hyprctl dispatch exec "hyprlock.sh cava  && rm /tmp/hyprlock-cava" ;} 
$WEATHER_CMD = echo "<b>Feels like<big> $(curl -s 'wttr.in?format=%t' | tr -d '+') </big></b>"
$KEYBOARD_LAYOUT = grep -m 1 "$(hyprctl -j devices | jq -r '.keyboards[] | select(.main == true) | .active_keymap')" /usr/share/X11/xkb/rules/base.lst | awk  '{print $1}'
$BATTERY_ICON = battery.sh icon

# hyprlang noerror true
# Source users custom hyprlock layout
source = $XDG_CONFIG_HOME/hypr/hyde.conf #
# hyprlang noerror false


# See the file first to see the equivalent colors
source = $XDG_CONFIG_HOME/hypr/themes/colors.conf

general{
    fractional_scaling=1
    # immediate_render=true
    # grace = true
}

background {
    color = $wallbash_pry1_rgba
}

label {
    text = cmd[update:86400000] font.sh resolve "$LAYOUT_PATH" & disown
    position = 50%, -20%
    font_size = 20
    valign = top
}


[...]

PR #845, opened 2025-08-05
core: monitor replug workaround for nvidia

PaideiaDilemma:
So I have been investigating #793 (Which has a confusing title, but I think all reports are caused by monitor re-plugging). Edit: wrong issue. I wanted to link this one: #695. They might be the same though. I think both probably have nothing to do with suspend per se.

I was able to reproduce the issue on this system:
```
Hyprland 0.50.0 built from branch main at commit 2859f1b795e1e772e9fc2132708ae03cd23ca39b  (keybinds: use the triggering keyboard for repeat timings (11309)).
Date: Tue Aug 5 15:54:55 2025
Tag: v0.50.0-68-g2859f1b7, commits: 6347
built against:
 aquamarine 0.9.2
 hyprlang 0.6.3
 hyprutils 0.8.2
 hyprcursor 0.1.13
 hyprgraphics 0.1.5


no flags were set


System Information:
System name: Linux
Node name: desktop
Release: 6.15.8-200.fc42.x86_64
Version: #1 SMP PREEMPT_DYNAMIC Thu Jul 24 13:26:52 UTC 2025


GPU information: 
0a:00.0 VGA compatible controller [0300]: NVIDIA Corporation GP102 [GeForce GTX 1080 Ti] [10de:1b06] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX x86_64 Kernel Module  575.64.05  Fri Jul 18 16:01:21 UTC 2025


os-release: NAME="Fedora Linux"
VERSION="42 (Workstation Edition)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=42
VERSION_CODENAME=""
PLATFORM_ID="platform:f42"
PRETTY_NAME="Fedora Linux 42 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:42"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f42/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=42
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=42
SUPPORT_END=2026-05-13
VARIANT="Workstation Edition"
VARIANT_ID=workstation


plugins:

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel DP-2: 3840x2160, DP-2 Samsung Electric Company U28E850 HTPK600065 -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ❌
		non-desktop ❌
		

```

I can get back to it tomorrow and hopefully I will find a fix for the issue.

Somehow we are failing to create and render to a new session lock surface after the previous one has been destroyed.

This issue does not occur with swaylock and is reproducible in hyprland and sway.

### DIFF
--- a/src/core/Egl.cpp
+++ b/src/core/Egl.cpp
@@ -38,8 +38,9 @@ CEGL::CEGL(wl_display* display) {
     if (eglCreatePlatformWindowSurfaceEXT == nullptr)
         throw std::runtime_error("Failed to get eglCreatePlatformWindowSurfaceEXT");
 
-    eglDisplay     = eglGetPlatformDisplayEXT(EGL_PLATFORM_WAYLAND_EXT, display, nullptr);
-    EGLint matched = 0;
+    const char* vendorString = nullptr;
+    eglDisplay               = eglGetPlatformDisplayEXT(EGL_PLATFORM_WAYLAND_EXT, display, nullptr);
+    EGLint matched           = 0;
     if (eglDisplay == EGL_NO_DISPLAY) {
         Debug::log(CRIT, "Failed to create EGL display");
         goto error;
@@ -64,6 +65,9 @@ CEGL::CEGL(wl_display* display) {
         Debug::log(CRIT, "Failed to create EGL context");
         goto error;
     }
+
+    vendorString = eglQueryString(eglDisplay, EGL_VENDOR);
+    m_isNvidia   = (vendorString) ? std::string{vendorString}.contains("NVIDIA") : false;
 
     return;
 
--- a/src/core/Egl.hpp
+++ b/src/core/Egl.hpp
@@ -17,6 +17,8 @@ class CEGL {
     PFNEGLCREATEPLATFORMWINDOWSURFACEEXTPROC eglCreatePlatformWindowSurfaceEXT;
 
     void                                     makeCurrent(EGLSurface surf);
+
+    bool                                     m_isNvidia = false;
 };
 
 inline UP<CEGL> g_pEGL;
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -1,5 +1,4 @@
 #include "hyprlock.hpp"
-#include "AnimationManager.hpp"
 #include "../helpers/Log.hpp"
 #include "../config/ConfigManager.hpp"
 #include "../renderer/Renderer.hpp"
@@ -18,8 +17,6 @@
 #include <cassert>
 #include <cstring>
 #include <xf86drm.h>
-#include <filesystem>
-#include <fstream>
 #include <algorithm>
 #include <sdbus-c++/sdbus-c++.h>
 #include <hyprutils/os/Process.hpp>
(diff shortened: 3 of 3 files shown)


PaideiaDilemma:
session lock surface never gets mapped.

We do successfully create a new EGLSurface and then proceed to call `eglSwapBuffers` after the reconnect, but the wayland debug log from hyprlock shows there is no `wl_surface.attach` or `wl_surface.commit`.

So it is some egl related issue. I tried to recreate the entire egl context on monitor disconnect, but that also didn't solve it.

PaideiaDilemma:
`eglSwapBuffers` returns `EGL_FALSE`

PaideiaDilemma:
Hmm didnt manage to solve it. I think its a NVIDIA/egl-wayland issue, but i am not sure. (I found some refcount issue in the eglSwapBuffer implementation there, but might be unrelated)

I cant get back to the machine i can reproduced it on for two weeks, so sadly this is all i got for now.

PaideiaDilemma:
Ok so this PR fixes the issue now.

Not sure if I mentioned it: It only happens when the last monitor is disconnected and reconnected.
Having an additional monitor that is not turned off magically makes it a non-issue. I think that hints towards it actually being an issue in the nvidia egl wayland stack.
Tested with nvidia 580.76.05 (lastest at time of writing this).
 
The fix is done by recreating the rendering context when the last monitor gets disconnected. This way we don't get `BAD_SURFACE` on `eglSwapBuffers` after creating a clean new surface+eglWindow+eglSurface and finally lockSurface for the output that gets readded. 
I think we clean up everything correctly on the disconnect (verified via printf debugging that its not an issue with some SP still having a reference, verified the locksurface gets deleted on the compositor side;).
I also spent quite a bit time trying to find another way to fix it and verifying that we are using the APIs correctly , but I didn't find anything.

vaxerski:
I mean if it works it works - but the context should absolutely not be lost when all monitors die. If it is, that's a driver bug.

You ideally would just reuse your current context to take a screenshot once a new monitor shows up

PaideiaDilemma:
I made this specifically for nvidia. We will find out if it also fixes @w1zpony's issue in #779 hopefully.
If so, this is not nvidia exclusive and we are doing something wrong.

I will still merge this as is for nvidia only because it is definitely an improvement. But I will try to stay on this issue as I want to get rid of this workaround at some point. Cheers!

MalteKW:
Fixed it for me, thank you so much.

PR #867, opened 2025-09-10
output: send wl_output release event

PaideiaDilemma:
This has the potential to fix some dynamic monitor issues.

I didn't try this, but in my mind it has the potential the monitor re-plugging issues on nvidia+.
Should be the correct thing to do anyways.


PaideiaDilemma:
Nevermind, I just realized that this is already part of the generated wayland.cpp.

```
CCWlOutput::~CCWlOutput() {
    if (!destroyed)
        sendRelease();
}
```

PR #877, opened 2025-09-17
core: discard monitor screenshot when it's removed

PaideiaDilemma:
I am sort of tempted to just merge this because I feel like this is what causes most of the monitor issues that are currently floating around. 
But I am not actually sure and it does make the screenshots not reusable after a monitor disconnect like mentioned below.
So I am hoping to get some testing on this PR. If it fixes it we can do a proper fix and try to preserve the screenshot frames.

**Description**:

Succeeds the nvidia workaround, which didn't work as intended. What actually fixed it is discarding the screenshot frame I think.

This change will be made obsolete when migrating to the new gatherer from hyprgraphics. Migrating to it will make us revision screenshot frame management anyways.

It does introduce a cosmetic problem however:
When a monitor gets disconnected, its screenshot can't be used anymore. I hope to fix this by rendering it into a framebuffer.

### DIFF
--- a/src/renderer/widgets/Background.cpp
+++ b/src/renderer/widgets/Background.cpp
@@ -7,6 +7,7 @@
 #include "../../helpers/MiscFunctions.hpp"
 #include "../../core/AnimationManager.hpp"
 #include "../../config/ConfigManager.hpp"
+#include "src/renderer/AsyncResourceGatherer.hpp"
 #include <chrono>
 #include <hyprlang.hpp>
 #include <filesystem>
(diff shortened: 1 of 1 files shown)


erikh:
Hey, if you can give me a week or so, I can probably test this patch out this weekend.

PaideiaDilemma:
Ok so i changed this, since this doesn't fix the nvidia issue.
However, currently on the nvidia workaround when we reconnect the monitor, the screenshot frame is just black.
This makes more sense and could potentially still fix something.

I am leaving this open for people to test.

PR #879, opened 2025-09-25
core: migrate to Hyprutils::CAsyncResourceGatherer and improve resource handling

PointerDilemma:
This PR utilizes `Hyprutils::CAsyncResourceGatherer` (https://github.com/hyprwm/hyprgraphics/pull/36) and adds some hyprlock specific stuff on top.

Because of this rewrite I decided to make two significant changes concerning the interface between the gatherer and the widgets:

## Resource ids and avoiding duplicate resources

Resource ids are not strings anymore, but instead they are a hash of what was requested.
Before the caller (usually widgets) had to come up with a unique string. Now the resource id is returned when requesting a resource.

Together with the fact that the resources now have a ref count, we avoid having duplicate resources in the resource manager. This is nice to have because when `monitor=""` there is one widget per monitor that needs the same resource.

## `onAssetUpdate` virtual method for widgets

Before we had to capture a weak reference to a widget in a lambda to safely call the asset callbacks.
Now we don't store a lambda as a callback, but only the weak widget reference. When the resource is finished, we call the virtual `onAssetUpdate` method on it.

This is implemented using a vector to allow for multiple widget callbacks for a single resource.

### DIFF
--- a/src/renderer/AsyncResourceManager.cpp
+++ b/src/renderer/AsyncResourceManager.cpp
@@ -248,7 +248,6 @@ bool CAsyncResourceManager::request(ResourceID id, const AWP<IWidget>& widget) {
     if (m_assets[id].texture) {
         // Asset already present. Dispatch the asset callback function.
         const auto& TEXTURE = m_assets[id].texture;
-        Debug::log(LOG, "onAssetUpdate {}", id);
         if (widget)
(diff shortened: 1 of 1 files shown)


PaideiaDilemma:
There it is. Currently still testing it too.

vaxerski:
in general looks fine, just needs testing

PR #899, opened 2025-10-19
fix: don't fade in while suspending

BlueZeeKing:
Currently, when suspending the system, hyprlock stills tries to fade in, so it
is still fading in when the system resumes. This change simply prevents the fade
in if the system is actively suspending.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,6 +5,7 @@
 #include "core/AnimationManager.hpp"
 #include <cstddef>
 #include <string_view>
+#include <sdbus-c++/sdbus-c++.h>
 
 void help() {
     std::println("Usage: hyprlock [options]\n\n"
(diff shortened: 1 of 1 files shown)


PaideiaDilemma:
This makes us block on dbus each time we start without allowing a configuration for that.
I am sort of against that, but would have nothing against it with a configuration option that is off by default.

I just don't understand why you can't launch hyprlock with `--no-fade-in` in whatever mechanism you use to launch hyprlock before suspend. In hypridle that would be `general:before_sleep_cmd = hyprlock --no-fade-in`.
That has a lot less overhead.

I feel like most of why people don't want to do that is that they for some reason want to use `loginctl lock-session` on suspend? I understand that the login dbus interface has a locked state, but one can still set that and still use --no-fade-in directly.

PR #903, opened 2025-10-22
background: reload screenshot if reload_cmd specified

Nathanwoodburn:
Allow setting a reload command for background when using screenshot.

One benefit of adding this featue is allowing locking with a screenshot while in the grace period and then switch to a different image.

### DIFF
--- a/src/renderer/widgets/Background.cpp
+++ b/src/renderer/widgets/Background.cpp
@@ -72,12 +72,13 @@ void CBackground::configure(const std::unordered_map<std::string, std::any>& pro
     } else if (!path.empty())
         resourceID = g_asyncResourceManager->requestImage(path, m_imageRevision, nullptr);
 
-    if (!isScreenshot && reloadTime > -1) {
+    if (!reloadCommand.empty() && reloadTime > -1) {
         try {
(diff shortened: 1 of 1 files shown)


src/renderer/widgets/Background.cpp
```diff
@@ -72,12 +72,14 @@ void CBackground::configure(const std::unordered_map<std::string, std::any>& pro
     } else if (!path.empty())
         resourceID = g_asyncResourceManager->requestImage(path, m_imageRevision, nullptr);
 
-    if (!isScreenshot && reloadTime > -1) {
+    // Allow reloads for screenshots if a reload command is provided
+    if ((!isScreenshot || !reloadCommand.empty()) && reloadTime > -1) {
```
PaideiaDilemma: we don't need to check if isScreenshot then. just `!reloadCommand.empty() && reloadTime > -1`

PaideiaDilemma (changes requested):

src/renderer/widgets/Background.cpp
Nathanwoodburn: Just pushed an amended commit to use this new if condition

PR #906, opened 2025-10-29
Add fragment shader widget

### LINKED ISSUE #133: [REQ] Fragment Shaders support
Would be awesome to have fragment shaders supports as in hyprland.
We could design some funky screensavers.

mashaal:
Allows a new shader widget config object: `shader {}`.

`monitor` and `zindex` are taken from the other config manager defaults - only `frag_path` is new and required. 

```ini
shader {
  frag_path = /path/to/shader.frag
  # monitor // optional
  # zindex // optional
}
```

This adds a new layer to hyprlock, works well with a transparent background to make some pretty neat looking overlays as your existing background config should still remain intact.

Ready (and willing) to do whatever it takes to get this merged. Closes #133 

https://github.com/user-attachments/assets/feda9d4e-e75c-43d4-b684-1ae07cfe2782

https://github.com/user-attachments/assets/292fde01-b1ab-4218-bcf9-77f7d9e53e4a

### DIFF
--- a/src/config/ConfigManager.cpp
+++ b/src/config/ConfigManager.cpp
@@ -326,6 +326,13 @@ void CConfigManager::init() {
     SHADOWABLE("label");
     CLICKABLE("label");
 
+    // shader widget for custom fragment shader backgrounds
+    m_config.addSpecialCategory("shader", Hyprlang::SSpecialCategoryOptions{.key = nullptr, .anonymousKeyBased = true});
+    m_config.addSpecialConfigValue("shader", "monitor", Hyprlang::STRING{""});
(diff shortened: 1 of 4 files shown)


mashaal:
A copy of the globe shader in the second video, I didn't think it should be committed as part of PR.

https://gist.github.com/mashaal/0c352c1d92767de0e4b804451c29f5ae

vaxerski:
this is DOA as we are planning to move to hyprtoolkit which will remove any shader functionality as toolkit doesn't expose the rendering backend.

It might be possible to add this in the future bot IMHO this is a dumb gimmick that we don't need in the project.

PR #908, opened 2025-11-04
Add support for mixed percentage/pixel in one axis of a layout value

Shiphan:
This add support for this syntax: `position = 5% + 122, 40%`, useful for placing multiple widgets with a constant gap to a relative position

lock


vaxerski (commented):
at this point please do not repeat Hyprland's mistakes and just include an expression parser like e.g. https://github.com/ArashPartow/exprtk

Choice of an engine is up to you, I didn't do the research, just please justify.

PR #917, opened 2025-11-17
auth: reworked pam auth, added pam_setcred

xJayMorex:
Without calling pam_setcred, unlocking the machine doesn't also unlock GPG based secrets. Reworked CPam::auth to make it less repetitive.


PointerDilemma:
Ok I mean I guess doing `pam_setcred` is fine.

But why move the handle to the class. I think it was totally fine before no?

Issue #803, opened 2025-06-16
Ignore Specific Keys in Password Input


Labhund:
### Description

**Is your feature request related to a problem? Please describe.**
When waking the lock screen, I often press the spacebar to wake the display. However, this adds a space to the password input, which is never needed and can cause login failures.

**Describe the solution you'd like**
- An option to ignore specific keys (such as the spacebar) in the password input field.
- Optionally, display a "badinput" message or visual feedback when such keys are pressed.

**Describe alternatives you've considered**
Currently, I have to use the mouse to wake the screen or manually delete spaces, which is inconvenient.

**Additional context**
This would improve usability for users who habitually press spacebar to wake the screen.

PR #933, opened 2025-12-21
input: allow ignoring space in password input

zcag:
Adds the boolean config option "prevent_space_input". 
When enabled, it prevents space character from being entered into the input field. 

Need for this also mentioned in https://github.com/hyprwm/hyprlock/issues/803


PointerDilemma (commented):
Uhm

zcag:
> Uhm

Hey, noticed that a commit from my own version of the build got into this pr as well (the one that autosubmits on 4 chars), took it off from my fork’s main now, so only the space input thing remains

PR #950, opened 2026-01-18
background: run reload_cmd in configure if path is empty

nikitax44:
Allow dynamically setting initial background image.


PointerDilemma (approved):
I didn't add the reload command, but for me this seems acceptable.

PR #962, opened 2026-02-20
fix(fingerprint): prevent sensor timeout on long-running sessions

JeremieRodon:
Introduces an inactivity timeout mechanism that pauses fingerprint verification after a configurable period (**default: 0 second/disabled**) when no user input is detected. This prevents some device drivers from disconnecting or disabling the sensor during extended lock sessions.

The verification automatically resumes when user activity is detected (mouse movement, clicks, or keyboard input). The sensor is properly released when paused and re-claimed when resuming.

May fixe #702 though that's unclear as the issue is not well documented

See #955 for more context and history on this PR.

### DIFF
--- a/src/auth/Fingerprint.cpp
+++ b/src/auth/Fingerprint.cpp
@@ -40,6 +40,8 @@ CFingerprint::CFingerprint() {
     m_sFingerprintReady                  = *FINGERPRINTREADY;
     static const auto FINGERPRINTPRESENT = g_pConfigManager->getValue<Hyprlang::STRING>("auth:fingerprint:present_message");
     m_sFingerprintPresent                = *FINGERPRINTPRESENT;
+    static const auto INACTIVETIMEOUT    = g_pConfigManager->getValue<Hyprlang::INT>("auth:fingerprint:inactive_timeout");
+    m_sInactiveTimeout = *INACTIVETIMEOUT;
 }
 
 CFingerprint::~CFingerprint() {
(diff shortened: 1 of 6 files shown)


kylemanna:
Been running this for a few days on Lenovo Slim 7i with flaky `goodixtls-55x4` firmware.

Previously it'd pretty much always display a message about timing out, with this it seems to recover most of the time. I suspect the other issues are due to the firmware :-/

After varying amounts of time (1 min - 25 minutes) `fprintd` disconnects leading to this issue from my initial observations. I set the inactive timeout to 15 seconds to disable the finger print sensor just before `hypridle` turns off the display again and this works better then ever.

PointerDilemma (changes requested):

JeremieRodon:
I took the liberty to rebase the branch on [0.9.5](https://github.com/hyprwm/hyprlock/commit/d75e93f8ee1721d70549d96f4d14bf2948aab70c) and I changed the 2 log lines to use the hyprutils logger helper.

PointerDilemma:
@JeremieRodon needs a clang-format. Otherwise looks ok.

However, I wonder if we should add idle/resume to the auth interface and use the idle-notify wayland protocol to trigger those methods.
The benefit of that would be that one, I think the lid-switch counts as an idle event on the compositor level and that would allow a user to open the laptop and use the fingerprint sensor right away; two, some pam usages might also benefit from such and idle mechanism.

JeremieRodon:
@PointerDilemma 
> needs a clang-format.

Oups. Done.

> I wonder if we should add idle/resume to the auth interface and use the idle-notify wayland protocol to trigger those methods.

Honestly, that sounds like the objectively correct solution. It seems obvious I should use existing idle-detection logic rather than stitching together a partial, custom one. But I don't have the slightest idea how to even begin to wire that up. I don't see anything related to idle-notify in the current codebase so I suppose that it is a more involved change and it far exceeds my C++ capabilities ️☹️ On the other hand, if it is just a matter of including the correct header and using an existing object, point me in the correct direction and I can give it a try.

PR #983, opened 2026-04-04
Feature/video background

mcgi5sr2:
given this another stab, hopefully this will make it easier to merge in future


mcgi5sr2:
build failure is just lack of ffmpeg dependency, I have two commits that will resolve it
1. just add ffmpeg dependency to nix
2. make it all optional at compile time

mcgi5sr2:
apologies for commit spam, the change in log::log arrived mid work on this.

https://github.com/user-attachments/assets/00338bc7-9c0d-44bc-bbfa-731dba98fd59

video of what this allows, as eye candy

PointerDilemma:
I have nothing against it in particular. But I have been telling people to use this method with Hyprland, which allows for this feature without support in hyprlock:

1. `hyprctl keyword misc:session_lock_xray 1`
2. make hyprlock background transparent 
```diff --git a/assets/example.conf b/assets/example.conf
index 2e28cb5a2e..5807e4ea04 100644
--- a/assets/example.conf
+++ b/assets/example.conf
@@ -35,8 +35,9 @@
 
 background {
     monitor =
-    path = screenshot
-    blur_passes = 3
+    path =
+    blur_passes = 0
+    color = rgba(00000000)
 }
 
 input-field {
```
3. launch smth like mpvpaper to run the video. `mpvpaper -o "loop" -l overlay DP-3 <video>`

The PR has some minor style issues, e.g. Single line if's, too much code in the Background widget. I am also a bit worried about memory safety in the decode thread. Otherwise it looks ok from a first glance.

mcgi5sr2:
yea I appreciate that works, but it didn't seem to me to be the way a lock screen should work with a transparent background and then requiring another programs overlay. I can keep this maintained for personal use, and if you want it upstream I am happy to push

Keyaruga-Reese:
I did build this version and is working great. I have been using this for over a week and haven't run into any issue.
I do hope, and pray that this gets merged. Or at least,  maintained as a fork.

Edit 1:
I did try making the hyprlock background transparent but I am not a fan of it. Moreover, mpvpaper (mpv in general) causes huge memory leaks on nvidia (i run hyprland on nvidia). 
Hence, in my case (may apply to many as there are people who use nvidia gpu's and run hyprland), I think this feature is pretty good.

mcgi5sr2:
I'll maintain it for the foreseeable, but as we all know that's not forever as I get into other work streams

mcgi5sr2:
@PointerDilemma  I've reworked the decode thread to tighten this up (0040ce5, 4008cda).

The decode thread owns all FFmpeg state after open() returns, the render thread owns the GL texture. The only shared state is a pre-sized buffer and a flag behind a mutex, published by std::swap, so no pointers into AVFrame data ever cross threads and nothing allocates after open(). The sws_scale destination is a fixed 4WH buffer with matching stride, so a mid-stream resolution change scales into the same bounds instead of overrunning them.

The thread also can't outlive the object: stop() sets an atomic flag, wakes the pacing wait, and joins before freeing anything. An interrupt callback on the format context aborts a blocked av_read_frame, so the join is bounded.

Style points are addressed too: video code is out of the Background widget, single-line ifs are gone, and the branch is clang-formatted. Also fixed since your review: BT.709 colourspace, display-matrix rotation, ~ path expansion, and cross-fade from the screenshot on fade-in.

I plan to look at the N monitor issue next, and even if we could move a lot of this into hypr-graphics and then have it as a call here, and perhaps hypr-paper as well.

mcgi5sr2:
used a Claude agent to review the PR found a couple of issues I agreed with so applied the later couple of commits.

mcgi5sr2:
Feature
- Video file backgrounds via FFmpeg (mp4, mkv, webm, mov, gif and others) with seamless looping
- Optional at compile time (VIDEO_BACKEND cmake flag, default on). Without FFmpeg the widgets compile against a stub and video paths are removed by the compiler
- Requires FFmpeg 6.1 or newer, checked at configure time

Correctness
- Files are opened on the decode thread, so a dead mount or FIFO cannot freeze the lock screen
- Colours converted with the stream's real colorspace (BT.601/709/2020) instead of swscale defaults
- Display matrix rotation honoured, including through the blur path, verified against VLC
- Frame timing uses stream start_time, fixing MPEG-TS captures that played fast or froze
- Decoder drained at end of file before looping, so B-frame clips loop cleanly and very short clips display at all
- Transparent video and gifs premultiplied and composited over the background colour
- Paths with ~ or relative components resolved before opening
- reload_cmd works with videos and can switch between image and video on reload
- Screenshot crossfade on lock fade-in now works for video backgrounds

Performance
- Redraws are driven by decoded frames, not display refresh: a 24fps video renders 24 times a second on any monitor
- Only the output showing the video is redrawn on a new frame
- Frames convert at no more than the output size, cutting bandwidth roughly 4x for 4K sources on smaller screens
- Static gifs and streams without timestamps pace at the container frame rate instead of spinning a core
- Unlock is immediate: the decode thread stops via interrupt callback and condition variable, no multi-second stall

Known limits
- Software decode only for now (36% of one core for 1080p24 on this machine, hardware decode blocked on Asahi AVD driver and FFmpeg support)
- One decoder per monitor showing the same file, shared decoding is planned as follow-up work
- .gif files now route to FFmpeg rather than the image loader

Next
- hardware decode
- N monitors shared decoder

Issue #986, opened 2026-04-10
SIGABRT in CSessionLockSurface::configure under GPU VRAM pressure (NVIDIA RTX 4090)


lvnilesh:
### Regression?

Unknown

### Hyprlock Info and Version

Hyprlock version v0.9.3

### Compositor Info and Version

Hyprland 0.54.0 (commit aaa2fc342f002bf4acd965f1ad2ead3796347e35)

### Bug description

hyprlock crashes with SIGABRT in `CSessionLockSurface::configure()` when the GPU is under VRAM pressure. The crash occurred 5 times in ~13 minutes.

**Trigger:** An application (llama-server) was consuming ~20.8 GB of the 23 GB VRAM on an RTX 4090. The NVIDIA driver was already failing GEM allocations:

```
[drm:nv_drm_gem_alloc_nvkms_memory_ioctl [nvidia_drm]] *ERROR* Failed to allocate NVKMS memory for GEM object
[drm] [nvidia-drm] Failed to allocate NVKMS video memory for GEM object, trying to fall back to sysmem
```

When hyprlock tried to lock the screen in this state, it hit an assertion/abort in `CSessionLockSurface::configure()` rather than failing gracefully.

### Stack trace

From `coredumpctl info`:

```
Signal: 6 (ABRT)

#0  __pthread_kill_implementation (libc.so.6)
#1  raise (libc.so.6)
#2  abort (libc.so.6)
#3  CSessionLockSurface::configure(Vector2D, serial).cold (hyprlock + 0x20fe7)
#4  std::_Function_handler<...CExtSessionLockSurfaceV1Configure...>::_M_invoke (hyprlock + 0x73e24)
#5  _CExtSessionLockSurfaceV1Configure (hyprlock + 0xe52e4)
...
#12 CHyprlock::run() (hyprlock + 0x86bc7)
#13 main (hyprlock + 0x33928)
```

### Expected behavior

hyprlock should handle GPU allocation failures gracefully -- either retry, fall back to a simpler rendering path, or display an error -- rather than aborting. The `.cold` path in `configure()` suggests an assertion that could be replaced with error handling.

### Environment

- **GPU:** NVIDIA GeForce RTX 4090 (24 GB VRAM)
- **Driver:** NVIDIA 595.58.03
- **Kernel:** 6.18.20
- **OS:** NixOS (nixos-unstable)
- **Hyprland:** 0.54.0
- **hyprlock:** 0.9.3

### Coredumps

[...]

PR #987, opened 2026-04-10
fix: handle EGL surface creation failure gracefully instead of aborting

lvnilesh:
## Summary

- Replace `RASSERT` (which calls `std::abort()`) with error logging and early return when `wl_egl_window_create` or `eglCreatePlatformWindowSurfaceEXT` fails in `CSessionLockSurface::configure()`
- Sets `readyForFrame = false` so the next configure event retries the allocation
- Allows hyprlock to survive transient GPU memory pressure instead of crashing and leaving the session in a broken lock state

## Problem

When GPU VRAM is exhausted (e.g., a large LLM loaded in VRAM), EGL surface allocation fails and `RASSERT` calls `std::abort()`, producing a SIGABRT crash. This leaves Hyprland displaying the "lockscreen app died" message with no way to recover without restarting the compositor.

See #986 for full crash details, stack traces, and coredumps.

## Test plan

- [ ] Exhaust GPU VRAM (e.g. allocate >90% with another process)
- [ ] Trigger screen lock via hyprlock
- [ ] Verify hyprlock logs the error instead of crashing
- [ ] Free VRAM and verify hyprlock recovers on next configure event
- [ ] Verify normal lock/unlock works without VRAM pressure

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -84,13 +84,21 @@ void CSessionLockSurface::configure(const Vector2D& size_, uint32_t serial_) {
 
     if (!eglWindow) {
         eglWindow = wl_egl_window_create((wl_surface*)surface->resource(), size.x, size.y);
-        RASSERT(eglWindow, "Couldn't create eglWindow");
+        if (!eglWindow) {
+            Debug::log(ERR, "Couldn't create eglWindow (GPU memory pressure?), will retry on next configure");
+            readyForFrame = false;
+            return;
+        }
     } else
         wl_egl_window_resize(eglWindow, size.x, size.y, 0, 0);
 
     if (!eglSurface) {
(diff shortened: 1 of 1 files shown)


PointerDilemma:
Thanks!

I thought about doing this multiple times, but a few Hyprland versions ago it seemed to risky, because `misc:session_lock_xray` was essentially enabled by default.

What will happen with this PR in the failure scenario should be the following.
On startup:
1. We fail to commit a lock surface to an output (there is no retry mechanism)
2. After a slight delay the "lockscreen app died" image will render on that output. A solid color on other compositors.
3. The compositor will kick us as the lock screen application, because we failed to commit a lock surface after some amount of time.
4. hyprlock exits gracefully after receiving the finished event. (Which is better than an abort)

When a new output gets added at runtime (1) and (2) happens. After that nothing will happen. One should still be able to unlock even though there might not be any UI.

Considering that, I think this is reasonable.

This PR just needs to be updated to use the new `Log::logger->log` function and also maybe be a bit more granular. Maybe use eglGetError.

PointerDilemma:
I included your commit in #988.

PR #988, opened 2026-04-10
lock-surface: improved handling of egl window and surface creation

PointerDilemma:
Successor of #987

Thanks @lvnilesh

I revised the logging a bit and also updated the code path for when
configure is called again (e.g. on a scale update).  Just to make sure we only resize and don't associate an existing eglWindow with another eglSurface, which is invalid according to the docs for eglCreateWindowSurface.

I tried to provoke one of those errors, by allocating lots of gpu memory, but I wasn't able to make eglCreateWindowSurface fail. Compositor always froze before that.

From my comment in #987 I expect the following failure scenarios if either creating window or surface fails:

On startup:
1.    We fail to commit a lock surface to an output (there is no retry mechanism)
2.   After a slight delay the "lockscreen app died" image will render on that output. A solid color on other compositors.
3.    The compositor will kick us as the lock screen application, because we failed to commit a lock surface after some amount of time.
4.   hyprlock exits gracefully after receiving the finished event. (Which is better than an abort)

When a new output gets added at runtime (1) and (2) happens. After that nothing will happen. One should still be able to unlock even though there might not be any UI.

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -82,20 +82,32 @@ void CSessionLockSurface::configure(const Vector2D& size_, uint32_t serial_) {
 
     surface->sendDamageBuffer(0, 0, 0xFFFF, 0xFFFF);
 
-    if (!eglWindow) {
+    if (eglWindow && eglSurface) {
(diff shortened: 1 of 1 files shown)


vaxerski:
looks fine to me

Issue #982, opened 2026-04-02
Hitting enter not Generating a New * in password,


astraea-oss:
#### Current Behaviour
Hitting Enter generates an addition ofuscated character in the password field, for the small duration before login happens.

#### Aestetic change,  
some way to disable this, and some other form of visual feedback that you hit the enter key that doesn't imply the enter key is part of the password.

NikolasKekelak:
Hi my name is Nick, 

I would like to contibute to Hyprland ecosystem, after reading few issues this one seems doable to me. Can I take this issue under my wings?

PointerDilemma:
@NikolasKekelak there is a lot of issue tracker work to do. Fixing bugs some of the bugs that existed for a while will be hard, but I am sure you will find something by clicking through the issue tracker and making yourself familiar with the code. Other Hyprland ecosystem application would also benefit from work like that.

This issue however is weird and it's not actually a thing, at least not in the way it is described.
@astraea-oss your keyboard may be broken. Or maybe you use something special in your input stack? You could use `wev` and check the keycodes that are listed when you hit enter.

astraea-oss:
I don't currently have access to a computer with the ability to install hyprlock, after moving some stuff around so can't verify the issue I was having. If it's unique to me then you have my apologies. perhaps in a little while after a fresh install. I can verify if it was unique to that particular instance of my setup.

husshpup:
@astraea-oss I had this issue with the following `input` section in `hyprland.conf`:

    input = {
        kb_layout = us,fr
        kb_options = caps:shiftlock
    }

I have also remapped CapsLock in the virtual console. All this to address two broken Shift keys : ). Unsure but perhaps that caused this issue. My solution was to install [keyd](https://github.com/rvaiya/keyd), which remaps CapsLock at the kernel level.

PR #989, opened 2026-04-12
Fix Enter key adding extra character to password field

NikolasKekelak:
Fixed an issue, where control characters could accidentaly enter passBuffer.

Changed one condition for consistency with rest of the code

(m_sPasswordState.passBuffer.length()>0 -> !m_sPasswordState.passBuffer.empty()) 

(hopefully)Fixes issue #982

### DIFF
--- a/src/core/hyprlock.cpp
+++ b/src/core/hyprlock.cpp
@@ -614,6 +614,8 @@ void CHyprlock::onKey(uint32_t key, bool down) {
     renderAllOutputs();
 }
 
+
+
 void CHyprlock::handleKeySym(xkb_keysym_t sym, bool composed) {
     const auto SYM = sym;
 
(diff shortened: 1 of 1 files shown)


PointerDilemma:
Hi! Please look at the code. There is no way that `SYM == XKB_KEY_Return || SYM == XKB_KEY_KP_Enter` can be added to the password buffer. Those are all `else if` branches. 
The issue you referenced is poorly worded and I have no evidence that this can actually happen.

NikolasKekelak:
Morning, yeah I saw that in the code. I just thought maybe on certain systems, for some reason Enter or Return key adds extra character to keyboard buffer and hyprlock registers it.

Sorry about your time, next time I will test thing more about issue I will be taking.

PR #995, opened 2026-04-17
input-field: fix placeholder not updating correctly

PointerDilemma:
Bug introduced in #959, reference made the config text be overwritten.

Was my bad. I somehow forgot how reference work for a second.


rain-lua:
lmao. did not notice either. :100:

Rezwo:
happens to the best of us

Issue #926, opened 2025-12-10
hyprlock not respecting PAM ordering?


johnrichardrinehart:
### Regression?

No

### Hyprlock Info and Version

```
$ hyprlock --version                      
Hyprlock version v0.9.2
```

<details> <summary>Hyprlock config</summary>

```sh
# sample hyprlock.conf
# for more configuration options, refer https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock
#
# rendered text in all widgets supports pango markup (e.g. <b> or <i> tags)
# ref. https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock/#general-remarks
#
# shortcuts to clear password buffer: ESC, Ctrl+U, Ctrl+Backspace
#
# you can get started by copying this config to ~/.config/hypr/hyprlock.conf
#

$font = Monospace

general {
    hide_cursor = false
}

auth {
    pam {
        enabled = true
    }
}

animations {
    enabled = true
    bezier = linear, 1, 1, 0, 0
    animation = fadeIn, 1, 5, linear
    animation = fadeOut, 1, 5, linear
    animation = inputFieldDots, 1, 2, linear
}

background {
    monitor =
    path = screenshot
    blur_passes = 3
}

input-field {
    monitor =
    size = 20%, 5%
    outline_thickness = 3
    inner_color = rgba(0, 0, 0, 0.0) # no fill

    outer_color = rgba(33ccffee) rgba(00ff99ee) 45deg
    check_color = rgba(00ff99ee) rgba(ff6633ee) 120deg
    fail_color = rgba(ff6633ee) rgba(ff0066ee) 40deg

    font_color = rgb(143, 143, 143)
    fade_on_empty = false
    rounding = 15

    font_family = $font
    placeholder_text = Input password...
    fail_text = $PAMFAIL

    # uncomment to use a letter instead of a dot to indicate the typed password
    # dots_text_format = *
    # dots_size = 0.4
    dots_spacing = 0.3

    # uncomment to use an input indicator that does not show the password length (similar to swaylock's input indicator)
    # hide_input = true

    position = 0, -20
    halign = center
    valign = center
}

# TIME
label {
    monitor =
    text = $TIME # ref. https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock/#variable-substitution
    font_size = 90
    font_family = $font

    position = -30, 0
    halign = right
    valign = top
}

# DATE
label {
    monitor =
[...]

PR #1000, opened 2026-04-28
auth: start PAM authentication immediately without pre-collecting input

Meldrey:
## Summary

- Remove the hardcoded "Password: " prompt and `waitForInput()` call from `CPam::init()` so that `pam_authenticate()` runs immediately on lock
- Modify the PAM conversation callback to block for input on every prompt, since input is no longer pre-collected
- Non-interactive PAM modules (face recognition via howdy/pam_python, FIDO2 devices) now run before any user interaction is required

## Problem

`CPam::init()` blocks on `waitForInput()` before ever calling `pam_authenticate()`. This means non-interactive PAM modules can never execute until the user types something and presses Enter. Face recognition (howdy), FIDO2 hardware keys, and any other passwordless PAM module are effectively broken — they only fire *after* the user has already submitted input, defeating the purpose of passwordless authentication.

This was acknowledged in #926 and is related to the security concerns in #535.

## Solution

Start `pam_authenticate()` immediately when the lock screen appears. The PAM stack evaluates modules in order:

1. Lock screen appears → `pam_authenticate()` fires
2. Non-interactive modules (e.g. howdy as `sufficient`) run first — no PAM conversation needed
3. **If they succeed** → `PAM_SUCCESS` → screen unlocks instantly, no keypress
4. **If they fail** → PAM continues to the next module (e.g. `pam_unix`)
5. `pam_unix` triggers the `conv()` callback with a "Password:" prompt → blocks for user input → normal password flow

The `conv()` callback no longer skips `waitForInput()` for the "initial" prompt, since input is no longer pre-collected before PAM runs.

## What this enables

- **Face unlock** (howdy + pam_python): camera fires immediately on lock, face match unlocks without any keypress
- **FIDO2 hardware keys**: device activates immediately instead of requiring an Enter press first
- **Any non-interactive PAM module**: works as the PAM stack intended

## Test plan

- [ ] Normal password unlock still works (type password, press Enter)
- [ ] Face recognition via howdy unlocks without keypress when configured as `sufficient` in `/etc/pam.d/hyprlock`
- [ ] Failed face recognition falls through to password prompt correctly
- [ ] SIGUSR1 unlock still works
- [ ] Grace period unlock still works
- [ ] Multiple failed attempts show correct fail text
- [ ] Fedora-style double password prompts still work (conv blocks on each prompt)

## Diff

One file changed: `src/auth/Pam.cpp` — 14 insertions, 16 deletions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/auth/Pam.cpp
+++ b/src/auth/Pam.cpp
@@ -17,32 +17,28 @@
 int conv(int num_msg, const struct pam_message** msg, struct pam_response** resp, void* appdata_ptr) {
     const auto           CONVERSATIONSTATE = (CPam::SPamConversationState*)appdata_ptr;
     struct pam_response* pamReply          = (struct pam_response*)calloc(num_msg, sizeof(struct pam_response));
-    bool                 initialPrompt     = true;
 
     for (int i = 0; i < num_msg; ++i) {
         switch (msg[i]->msg_style) {
             case PAM_PROMPT_ECHO_OFF:
             case PAM_PROMPT_ECHO_ON: {
-                const auto PROMPT        = std::string(msg[i]->msg);
-                const auto PROMPTCHANGED = PROMPT != CONVERSATIONSTATE->prompt;
+                const auto PROMPT = std::string(msg[i]->msg);
                 Log::logger->log(Log::INFO, "PAM_PROMPT: {}", PROMPT);
 
-                if (PROMPTCHANGED)
-                    g_pHyprlock->enqueueForceUpdateTimers();
-
-                // Some pam configurations ask for the password twice for whatever reason (Fedora su for example)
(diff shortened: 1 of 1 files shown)


PointerDilemma:
This reverts https://github.com/hyprwm/hyprlock/pull/409

But also it adds a bunch of irrelevant comments and removes the workaround multiple prompts with the same prompt text that is there for fedora (idk if it still behaves that way, but it asked for "Password: " twice previously).

If you don't mind I will just revert that PR instead.

PointerDilemma:
Doesn't seem like that was tested at all. It is broken. `isUnlocked` is a sort of a bad interface and is not a valid check now that waitOnInput is not called right away.
 
Succeeded by #1003

Issue #953, opened 2026-01-26
Hyprlock/Hyprland crash investigation: multiple causes and fixes (DPMS, suspend, PAM)


brenoperucchi:
### System Info
- **Hyprlock:** 0.9.2
- **Hyprland:** 0.53.1
- **OS:** Arch Linux (kernel 6.18.3)
- **GPU:** AMD Radeon RX 470/480/570/580 (amdgpu driver)

### Summary

After weeks of investigating repeated Hyprland crashes during idle/lock states, I found **multiple independent issues** that can cause crashes. Documenting all of them here to help others.

---

## Issue 1: DPMS off crashes Hyprland (AMD GPU)

**Symptom:** Hyprland crashes when DPMS turns off monitors.

**Cause:** Known AMD GPU issue with DPMS in Hyprland.

**Fix:** Remove DPMS off from hypridle config.

**Related:** hyprwm/Hyprland#2578

---

## Issue 2: Monitor disable crashes Hyprlock

**Symptom:** Hyprlock crashes when monitors are disabled via `hyprctl dispatch dpms off` or monitor power management.

**Cause:** Hyprlock cannot handle displays becoming unavailable.

**Fix:** Do not disable monitors while hyprlock is running. Keep system in screensaver + lock only.

**Related:** #434, #480

---

## Issue 3: AMD GPU suspend issues

**Symptom:** System freezes or crashes on suspend/resume.

**Cause:** AMD GPU runtime power management issues.

**Fix:** Add kernel parameters:
```
amdgpu.runpm=0 mem_sleep_default=deep
```

---

## Issue 4: Fingerprint auth crash ("User unknown")

**Symptom:** Hyprlock crashes with "User unknown" PAM error.

**Cause:** `fingerprint:enabled = true` in hyprlock.conf but `fprintd` is not installed.

**Fix:** Disable fingerprint in hyprlock.conf:
```ini
# auth {
#     fingerprint:enabled = true
# }
```

---

## Issue 5: Missing PAM header

**Symptom:** PAM authentication fails silently.

**Cause:** `/etc/pam.d/hyprlock` missing `#%PAM-1.0` header.

**Fix:** Ensure file starts with `#%PAM-1.0`.

---

## Issue 6: Incomplete PAM configuration (pam_end crash) - NEW

**Symptom:** Hyprlock crashes with SIGABRT in `pam_end()`, then Hyprland crashes.

**Stack trace:**
```
[...]

PR #1001, opened 2026-04-28
core: fix freeze when a monitor is DPMS-off during lock

tmbull:
# Summary                  

Fix hyprlock freezing indefinitely when one monitor is DPMS-off at lock time, by preventing eglSwapBuffers from blocking on a compositor that will never release the buffer for a powered-off display.

**Caveat:** most of the work here was done by Claude. I am a developer, but I haven't done any C++ in ~15 years and don't have a good understanding of the internal workings of hyprland or display servers, etc. in general. However, I noticed there have been a lot of related issues posted, so I thought I would put this up for review by the maintainers. I'm happy to help further.

# Problem                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                       
This seems possibly related to #953, #695, and maybe others. I have an NVidia RTX 3060Ti, a Samsung G9 (ultrawide) monitor, and an older 24" Samsung monitor. I'm currently using the nvidia-open drivers (595.58.03) on nixos.

The exact behavior I was seeing:

* On transition to lock screen, the screen would freeze (I would see my session with a shadow of the lock screen over it).
* I could enter my password (blindly), but the screen would still be frozen until I manually powercycled my monitor
* After powercycling, the system would unlock
* After the freezing issue was fixed, I also noticed that hyprlock would segfault on unlock, so that is addressed here as well.

I tried setting `screencopy_mode = 1` to see if that had any effect, but it did not.
                                                                                                                                                                                                                                                                                         
 # Solution                                                                                                                                                                                                                                                                             

  Egl.cpp — Call eglSwapInterval(0) in `configure`. This makes the swap non-blocking: the buffer is committed to the compositor and the call returns immediately. Rendering rate is already properly throttled by wl_surface.frame callbacks, so this does not cause over-rendering.


  ~~LockSurface.cpp — When configure() acknowledges a new serial, reset any pending frame callback before calling render(). This ensures every ack_configure is followed by a committed frame, breaking the protocol deadlock during rapid output reconfiguration.~~
  ~~hyprlock.cpp — Skip zwp_linux_dmabuf_v1 binding when screencopy_mode=1. The GBM device created in the dmabuf feedback callback is only needed for DMABUF screencopy; creating it unnecessarily can block on the DRM render node under the same adverse conditions.~~
  ~~Output.cpp — Guard createSessionLockSurface() with m_bTerminate to prevent a segfault when an output reconnects during session unlock teardown (the session lock object is already destroyed at that point).~~

# Test steps
  - Normal password unlock works without monitor power-cycle
  - Fade-in and fade-out animations play correctly on all monitors
  - Lock/unlock works with a monitor in DPMS-off state at lock time 
  - ~~No segfault when a monitor reconnects during unlock~~                                                                                                                                                                                                                 
  - ~~screencopy_mode = 1 no longer attempts dmabuf setup~~
  - ~~Default DMABUF screencopy mode still works when monitors are active~~

🤖 Generated with Claude Code

### DIFF
--- a/src/core/LockSurface.cpp
+++ b/src/core/LockSurface.cpp
@@ -105,6 +105,13 @@ void CSessionLockSurface::configure(const Vector2D& size_, uint32_t serial_) {
             readyForFrame = false;
             return;
         }
+
+        // Disable vsync throttling so eglSwapBuffers never blocks waiting for
+        // wl_buffer.release. Without this, a DPMS-off monitor can stall the main
+        // thread indefinitely because the compositor never releases the buffer.
+        // Rendering rate is already governed by wl_surface.frame callbacks.
+        g_pEGL->makeCurrent(eglSurface);
+        eglSwapInterval(g_pEGL->eglDisplay, 0);
     }
 
     if (readyForFrame && !(SAMESIZE && SAMESCALE)) {


src/core/Egl.cpp
```diff
@@ -121,6 +121,12 @@ void CEGL::makeCurrent(EGLSurface surf) {
 }
 
 bool CEGL::swapBuffers(EGLSurface surf) {
+    // Disable vsync throttling so eglSwapBuffers never blocks waiting for
+    // wl_buffer.release. Without this, a DPMS-off monitor can stall the main
+    // thread indefinitely because the compositor never releases the buffer.
+    // Rendering rate is already governed by wl_surface.frame callbacks.
+    eglSwapInterval(eglDisplay, 0);
```
PointerDilemma: Yeah this seems ok to do just in case, but why do it before every swap? I dont think that is required.

src/core/LockSurface.cpp
```diff
@@ -78,6 +78,13 @@ void CSessionLockSurface::configure(const Vector2D& size_, uint32_t serial_) {
     if (!SAMESERIAL)
         lockSurface->sendAckConfigure(serial);
 
+    // If we acknowledged a new configure serial but a frame callback is still pending,
+    // the compositor may withhold the callback until we commit a frame for the new serial.
+    // Reset the callback so render() can proceed immediately and commit a fresh frame,
+    // preventing a deadlock when rapid configure events arrive (e.g. during DPMS transitions).
+    if (!SAMESERIAL && frameCallback)
+        frameCallback.reset();
```
PointerDilemma: This seems good. But like I also feel like the compositor should make sure this can't happen.

PointerDilemma (changes requested):

PointerDilemma:
Thanks a lot this is nice. Only thing really is that i think setting the swap interval once should be enough. Since the setting is bound to the surface, according to the docs, maybe call it in configure?

tmbull:
Thanks for the review! I made the requested change. I tested on my setup and it still works as expected. Please let me know if you'd like to make any other changes or test anything else in particular!

src/core/Egl.cpp
tmbull: I made the requested change, and it seems to be working as expected.

PointerDilemma:
But does it actually fix what you described?
I am a bit in doubt about that, since I experimented with `eglSwapInterval` Before and also with frame callback stuff.
Can you send some evidence like a recording maybe?

src/core/LockSurface.cpp
vaxerski: it can't because dpms is not a property exposed to clients

vaxerski:
generally this makes sense but I see three unrelated changes lole

src/core/LockSurface.cpp
PointerDilemma: I mean the comment doesn't make sense. I was thinking about a situation where a frame callback is active and configure is called again

tmbull:
Thanks for the comments. I  was experiencing two separate issues: one with the lock screen freezing while loading and another with the screen freezing after successfully entering my password. I'll get a video recorded to demo the fix, but it might take me a couple of days.

tmbull:
Ok, so I spent a bit more time on this today and tracked the actual fix down to just the `eglSwapInterval` change. The rest was just some hallucinating on the way to the actual solution.

@PointerDilemma I know you said you messed with `eglSwapInterval` previously, but maybe the issue I was experiencing is slightly different to the previously reported issue. 

Here are a couple of recordings. I apologize in advance for the quality.

You can see in the first that the screen freezes while transitioning to the lock screen. I click on the shadow of the password entry box, enter my password, powercycle the monitor, and hyprlock then exits successfully. 

[hyprlock g9 broken](https://youtu.be/y8zz-KGIQdw)

Here's a video after the fix: 

[hyprlock g9 fixed](https://youtu.be/IZ5s8nGhLBg)

I just enter my password and the screen unlocks as usual.

PointerDilemma (approved):

PointerDilemma:
Thanks!

Generally this is OK usage of AI, but I wanted to rewrite the comment, because nobody actually researched if the `wl_buffer.release` event is actually the mechanism used by the egl wayland implementation to do vsync with eglSwapInterval.
I didn't really like the comment also because the point is just that when we manage frame callbacks ourselves, we don't want to use any sync mechanism that comes with EGL.

https://github.com/hyprwm/hyprlock/pull/1015
I added you as a co-author, but omitted claude.

tmbull:
@PointerDilemma I totally understand. I debated contributing "vibe-coded" changes at all in a domain in which I have limited understanding, but I noticed there were several related issues. I'm just happy you were able to find value in the contribution.

PR #1010, opened 2026-05-09
auth: add Visage face authentication module

SomeCodecat:
Adds a parallel face-auth backend that talks directly to [Visage](https://github.com/sovren-software/visage)'s system D-Bus interface (`org.freedesktop.Visage1`) instead of going through `pam_visage.so`. This mirrors the approach taken for fingerprint in #514: PAM modules are evaluated as soon as `pam_authenticate` starts, which means the face check fires the instant the lock appears and unlocks immediately if a face is in front of the camera. Direct D-Bus integration lets us add a small `start_delay` after lock and gate retries on a condvar so submitting input nudges a fresh attempt.

A worker thread polls `Verify(username)` and on a match calls `g_pAuth->enqueueUnlock()`. PAM continues to run in parallel as before, so password unlock works unchanged.

Disabled by default. Enable with `auth.visage.enabled = 1`. Other knobs are `ready_message`, `retry_delay` (default 500 ms), and `start_delay` (default 2000 ms).

Verify calls are bounded with `withTimeout(30s)` so a hung daemon can't deadlock `terminate()`. On D-Bus errors the worker backs off exponentially up to 5 s and never permanently gives up — visaged commonly takes 30–90 s to come back after a hibernate resume on Surface devices, and the user may also be waiting on `visage-resume.service` to restart it. If visaged isn't running at hyprlock start, the worker retries the connection every 2 s.

I did not add a `PrepareForSleep` handler like fingerprint has. The retry loop recovers organically once visaged is reachable again, so it didn't seem worth the extra dbus connection. Happy to add it if you'd rather mirror the fingerprint structure exactly.

### DIFF
--- a/src/auth/Visage.cpp
+++ b/src/auth/Visage.cpp
@@ -1,0 +1,147 @@
+#include "Visage.hpp"
+#include "DBus.hpp"
+#include "../core/hyprlock.hpp"
+#include "../helpers/Log.hpp"
+#include "../config/ConfigManager.hpp"
+
+#include <chrono>
+#include <pwd.h>
+#include <unistd.h>
+
+// Per-call timeout on the Verify D-Bus method. visaged opens the camera, runs
+// inference, and replies (typically ~500-1500ms). After hibernate or on a
+// stale camera fd it can take far longer; cap so a stuck visaged can't
+// deadlock the dispatch.
+constexpr int VERIFY_TIMEOUT_MS = 30'000;
+
+// Backoff parameters when Verify keeps failing (visaged unavailable,
+// restarting, etc). delay = min(retry_delay << min(errors, MAX_BACKOFF_SHIFT),
+// MAX_BACKOFF_MS).
+constexpr int MAX_BACKOFF_MS    = 5'000;
+constexpr int MAX_BACKOFF_SHIFT = 4;
+
+static const auto VISAGE_SERVICE   = sdbus::ServiceName{"org.freedesktop.Visage1"};
+static const auto VISAGE_OBJECT    = sdbus::ObjectPath{"/org/freedesktop/Visage1"};
+static const auto VISAGE_INTERFACE = sdbus::InterfaceName{"org.freedesktop.Visage1"};
+static const auto LOGIN_MANAGER    = sdbus::ServiceName{"org.freedesktop.login1.Manager"};
+
+CVisage::CVisage() {
+    static const auto READYMSG = g_pConfigManager->getValue<Hyprlang::STRING>("auth:visage:ready_message");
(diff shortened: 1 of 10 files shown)


src/auth/Visage.hpp
```diff
@@ -0,0 +1,56 @@
...
+    std::optional<std::string> getLastPrompt() override;
+    void                       terminate() override;
+
+  private:
+    void workerLoop();
+    void interruptibleWait(int milliseconds);
+    bool setupProxy(std::shared_ptr<sdbus::IConnection>& outConnection, std::unique_ptr<sdbus::IProxy>& outProxy);
+
+    std::atomic<bool>       m_bAbort{false};
+    std::atomic<bool>       m_bAuthenticated{false};
```
PointerDilemma: This is the "legacy" way that hypr* named member vars. Can you change it to the new "m_<name>" format?

At some point I would like to bulk change all of the var names, but I don't want to disrupt open PRs...

src/auth/Visage.cpp
```diff
@@ -0,0 +1,153 @@
...
+        m_sUsername = pw->pw_name;
+}
+
+CVisage::~CVisage() {
+    terminate();
+}
+
+void CVisage::init() {
+    m_sPrompt = m_sReadyMessage;
+    m_thread  = std::thread([this] { workerLoop(); });
```
PointerDilemma: The fingerprint auth implementation doesn't need it's own thread, because it uses async method calls.
Is there a reason visage needs a thread?
I would prefer if visage would also handle it like that.

But for this to work, we need to pull out the dbus connection object from fingerprint (add `g_dbus`) so that we can reuse the same connection here and only need one connection.

PointerDilemma: Let me know if you would be up to pull out the dbus connection from fingerprint. If not I can do it and you can rebase your branch on top of it.

SomeCodecat:
Force-pushed `c3333ab` addressing both review threads:

  - Pulled the system bus connection out of `CFingerprint` into a shared `g_dbus`
  (`src/auth/DBus.{hpp,cpp}`). `hyprlock.cpp` polls it directly.
  - Rewrote Visage around `callMethodAsync` + `uponReplyInvoke` — no worker thread, no mutex/cv.
   Retries and the start delay go through `addTimer`; login1 `PrepareForSleep` is handled
  per-module on the shared connection. Added a one-line log in `handleInput` when the nudge path
   fires.
  - Renamed all Visage members to the new `m_<name>` form.

  Tested locally end-to-end (happy path, nudge-on-Enter, visaged stop/start, suspend/resume).

PR #1013, opened 2026-05-14
feat: add `success_text` for successful authentication

sseu-buhzzi:
Introduce `input-field:success_text`, `input-field:success_color`, and `general:success_timeout` to provide visual confirmation after successful authentication.

- Added `general:success_timeout` (default 0) to delay unlocking.
- Added `input-field:success_text` and `input-field:success_color` for custom feedback.

### DIFF
--- a/src/auth/Auth.cpp
+++ b/src/auth/Auth.cpp
@@ -76,12 +76,27 @@ void CAuth::terminate() {
     }
 }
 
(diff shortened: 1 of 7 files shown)


PointerDilemma (changes requested):
1. please call it "success_*"
2. just integrate the success_timeout and the flag with `unlockCallback`. No need to add another handler.

PR #1016, opened 2026-05-14
screencopy: flip the right 10bit formats

gulafaran:
seems the flip logic was flipping wrong formats
flip the right ones to fix red/blue swap on 10bit formats.

im not entirerly sure here asked AI even, seems very convulated flipping logic. but fixed the red/blue color swap on 10bit formats in hyprland when using screencopy_force_8b = false.


PointerDilemma:
Thanks! Just curious, dmabuf no worky for you?

gulafaran:
> Thanks! Just curious, dmabuf no worky for you?

no idea just stumbled upon this when doing https://github.com/hyprwm/Hyprland/pull/14543 and https://github.com/hyprwm/hyprgraphics/pull/46 and doing so saw the force 8bit format thing disabled it and noticed it color flipping

PR #1017, opened 2026-05-15
input-field: add auto_submit_after option

### LINKED ISSUE #1012: PIN auto-submit after N digits typed
### Description

It would be great to have an option in the `input-field` block that automatically submits the input after a specified number of characters are typed, similar to how Windows PIN login or phone lock screens work (no Enter key required). I am currently using a PIN on hyprlock via PAM with `pam_pwdfile`. This uses PAM to keep the PIN completely separate from the system password used for sudo, SSH, and login.

erasylurazalin:
Adds a new optional config key auto_submit_after to input-field. When set to N > 0, the input field automatically submits after N characters are typed without requiring Enter, similar to a phone PIN screen.

Submission is deferred by one frame so the final dot animates before PAM begins checking. A pending timer is cancelled if the buffer changes before it fires.

Fixes: https://github.com/hyprwm/hyprlock/issues/1012

### DIFF
--- a/src/renderer/Screencopy.cpp
+++ b/src/renderer/Screencopy.cpp
@@ -373,7 +373,7 @@ void CSCSHMFrame::convertBuffer() {
                 Log::logger->log(Log::INFO, "[sc] [shm] Converting 10-bit channels to 8-bit");
                 uint8_t*   data = (uint8_t*)m_shmData;
 
-                const bool FLIP = m_shmFmt != WL_SHM_FORMAT_XBGR2101010;
+                const bool FLIP = m_shmFmt == WL_SHM_FORMAT_XRGB2101010 || m_shmFmt == WL_SHM_FORMAT_ARGB2101010;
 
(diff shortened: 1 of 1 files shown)


PointerDilemma (changes requested):
No, this is the wrong way to implement it.

Since input handling is not tied to input-field in hyprlock (input-field is just a visualization of it), this feature should be in the `general` config section and should be triggered directly where chars are added to the password buffer. that way it's also a lot easier to implement.

PointerDilemma:
Also check my comment in #1012. I really don't want people to enable if they have a normal pam configuration tied to their user password. Any mechanism to make sure users are aware of the implications of this setting is good.

erasylurazalin:
Moved the logic to handleKeySym in the core as suggested. PasswordInputField is now untouched. Also added the warning comment next to the config registration about the password length leak. thanks for the feedback

PR #1026, opened 2026-05-21
pam: fix behaviour when pam_fprintd.so is used

PointerDilemma:
Setting m_bBlockInput to false on init was the actual fix.

Making sure that PAMPROMPT reflects pam_fprintd.so PAM_TEXT_INFO message should make it easier to see whats going on.

Using pam_fprintd.so works, but it's not parallel. Using the native fprintd support is adviced.


PointerDilemma:
Problem was introduced by #1003

PR #1027, opened 2026-05-22
base ubuntu dont exist libpam.pc only pam.pc

rsvzz:
In Ubuntu, a symbolic link from pam.pc to libpam.pc is needed to allow compiling, otherwise impossible

In Debian or Ubuntu-based distributions libpam.pc does not exist only pam.pc exists. 
The correction is sufficient to find a lib.

cmake allows compiling on Ubuntu distributions


rsvzz:
The problem occurs in Ubuntu 26.04 containers. This changed is needed.

vaxerski (approved):

PR #1028, opened 2026-05-28
fprint: introduce auth:fingerprint:max_attempts

PointerDilemma:
Successor of #833


filip-rs (approved):
nice, i also want this feature, lgtm

PR #1040, opened 2026-06-22
fix: use explicit cast for `m_pSeat`

onion108:
Fix build failure due to https://github.com/hyprwm/hyprutils/commit/9d8bf6e810597152eef8906c670b96679af2faec.

### DIFF
--- a/src/core/Seat.cpp
+++ b/src/core/Seat.cpp
@@ -169,5 +169,5 @@ void CSeatManager::registerCursorShape(SP<CCWpCursorShapeManagerV1> shape) {
 }
 
 bool CSeatManager::registered() {
-    return static_cast<bool>(m_pSeat);
+    return !!m_pSeat;
 }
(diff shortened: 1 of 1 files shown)


src/core/Seat.cpp
```diff
@@ -169,5 +169,5 @@ void CSeatManager::registerCursorShape(SP<CCWpCursorShapeManagerV1> shape) {
 }
 
 bool CSeatManager::registered() {
-    return m_pSeat;
+    return static_cast<bool>(m_pSeat);
```
vaxerski: we use `!!` instead.

onion108: ok i'll use that

vaxerski (approved):

tekstryder:
Thanks! Updated https://github.com/hyprwm/Hyprland/discussions/15201#discussioncomment-17391921.

PR #1059, opened 2026-08-06
pam: fix deadlock on terminate

### LINKED ISSUE #1055: Random system suspends, hibernates, and hyprlock freezes/segfaults on unlock (since v0.56)
### Regression?

Yes

### Hyprlock Info and Version

v0.9.6
<details> <summary>Hyprlock config</summary>

```sh
background {
  monitor =
  path = ~/dotfiles/wallpapers/night-sky.png
  color = rgba(25, 20, 20, 1.0)
  blur_passes = 0
}

input-field {
  size = 180, 60
  outline_thickness = 0
  placeholder_text = Input password...
  
  dots_size = 0.18
  dots_spacing = 0.5
  dots_center = true
  
  inner_color = rgb(E0E0E0)
  font_color = rgb(10, 10, 10)
  check_color = rgb(249, 225, 174)
  fail_color = rgb(255, 97, 97)
  
  hide_input = false
  rounding = -1

  fade_on_empty = true
  fade_timeout = 5000

  position = 0, -20
  halign = center
  valign = center
  font_family = JetBrainsMono Nerd Font
}

label {
  text = $TIME
  color = rgb(F2F2F2)
  font_size = 120
  font_family = JetBrainsMono Nerd Font
  position = 0, 190
  halign = center
  shadow_passes = 5
  shadow_size = 10
}
```

</details>

### Compositor Info and Version

<details> <summary>System/Version info</summary>

```sh
Hyprland 0.56.0 built from branch v0.56.0 at commit 36b2e0cfe0c6094dbc47bd42a437431315bb3087 clean (version: bump to 0.56.0).
Date: Mon Jul 20 07:50:48 2026
Tag: v0.56.0, commits: 7625

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.0, system has 0.14.0
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.13.0, system has 0.13.0

Version ABI string: 36b2e0cfe0c6094dbc47bd42a437431315bb3087_aq_0.13_hu_0.14_hg_0.5_hc_0.1_hlg_0.6
no flags were set


System Information:
System name: Linux
Node name: arch-nitrov15
Release: 7.1.4-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Sat, 18 Jul 2026 17:30:57 +0000

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.0, system has 0.14.0
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.13.0, system has 0.13.0



GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation Raptor Lake-P [UHD Graphics] [8086:a7a8] (rev 04) (prog-if 00 [VGA controller])
0000:01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD107M [GeForce RTX 4050 Max-Q / Mobile] [10de:28a1] (rev a1) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  610.43.03  Release Build  (root@)  


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo


plugins:

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel eDP-1: 1920x1080, eDP-1 BOE 0x0D39  -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ✔️
		non-desktop ❌
		

State:

configProvider: lua
backend: drm
```

</details>

### Description

I am using an Acer Nitro V15 with Hyprland, hyprlock, and hypridle. Recently, my system has started unexpectedly suspending/hibernating/going to sleep while in active use.

Furthermore, when trying to unlock the device after it wakes up, hyprlock frequently freezes or crashes (throwing a segmentation fault). The screen gets stuck indefinitely on the lock screen, forcing me to switch to a TTY to kill the process and recover my session.

Both of these issues started occurring after the **v0.56** update.

### How to reproduce

The unexpected suspensions happen randomly during normal usage, so I don't have exact steps to trigger them. However, the crash on unlock is very consistent:
 
1. Wait for the laptop to suspend (either randomly or manually).
2. Wake up the device.
3. Attempt to unlock via hyprlock.
4. Observe the freeze/crash (resulting in an infinite hang or a segfault requiring TTY recovery).

### Crash reports, logs, images, videos

I have attached the coredump.log generated via coredumpctl info hyprlock.
Note: The crash logged is a Signal 6 (ABRT), rather than a standard Signal 11 Segfault. I don't find where the log containing the seg fault error can be. I just see it in the TTY once.

[coredump.log](https://github.com/user-attachments/files/30364584/coredump.log)

vaxerski:
ref #1055

needs testing

### DIFF
--- a/src/auth/Pam.cpp
+++ b/src/auth/Pam.cpp
@@ -12,12 +12,25 @@
 #endif
 
 #include <cstring>
+#include <cstdlib>
 #include <thread>
 
 int conv(int num_msg, const struct pam_message** msg, struct pam_response** resp, void* appdata_ptr) {
     const auto           CONVERSATIONSTATE = (CPam::SPamConversationState*)appdata_ptr;
     struct pam_response* pamReply          = (struct pam_response*)calloc(num_msg, sizeof(struct pam_response));
     bool                 initialPrompt     = true;
+
+    *resp = nullptr;
+
+    if (!pamReply)
+        return PAM_BUF_ERR;
+
+    const auto abortConversation = [&]() {
+        for (int i = 0; i < num_msg; ++i)
+            free(pamReply[i].resp);
+        free(pamReply);
+        return PAM_CONV_ERR;
+    };
 
     for (int i = 0; i < num_msg; ++i) {
         switch (msg[i]->msg_style) {
@@ -37,8 +50,8 @@ int conv(int num_msg, const struct pam_message** msg, struct pam_response** resp
                 }
 
                 // Needed for unlocks via SIGUSR1
-                if (g_pHyprlock->isFadingOutOrTerminating())
-                    return PAM_CONV_ERR;
+                if (CONVERSATIONSTATE->terminateRequested || g_pHyprlock->isFadingOutOrTerminating())
+                    return abortConversation();
 
                 pamReply[i].resp = strdup(CONVERSATIONSTATE->input.c_str());
                 initialPrompt    = false;
@@ -78,7 +91,7 @@ CPam::CPam() {
 }
 
 CPam::~CPam() {
-    ;
+    terminate();
 }
 
 void CPam::init() {
@@ -87,13 +100,13 @@ void CPam::init() {
             resetConversation();
 
             // For grace or SIGUSR1 unlocks
-            if (g_pHyprlock->isFadingOutOrTerminating())
+            if (m_sConversationState.terminateRequested || g_pHyprlock->isFadingOutOrTerminating())
                 return;
 
             const auto AUTHENTICATED = auth();
 
             // For SIGUSR1 unlocks
-            if (g_pHyprlock->isFadingOutOrTerminating())
+            if (m_sConversationState.terminateRequested || g_pHyprlock->isFadingOutOrTerminating())
                 return;
 
             if (!AUTHENTICATED)
@@ -147,7 +160,8 @@ void CPam::waitForInput() {
     m_bBlockInput                          = false;
     m_sConversationState.waitingForPamAuth = false;
     m_sConversationState.inputRequested    = true;
-    m_sConversationState.inputSubmittedCondition.wait(lk, [this] { return !m_sConversationState.inputRequested || g_pHyprlock->isFadingOutOrTerminating(); });
+    m_sConversationState.inputSubmittedCondition.wait(
+        lk, [this] { return !m_sConversationState.inputRequested || m_sConversationState.terminateRequested || g_pHyprlock->isFadingOutOrTerminating(); });
     m_bBlockInput = true;
 }
 
@@ -176,6 +190,11 @@ bool CPam::checkWaiting() {
 }
 
 void CPam::terminate() {
+    {
+        std::lock_guard<std::mutex> lg(m_sConversationState.inputMutex);
+        m_sConversationState.terminateRequested = true;
+    }
+
     m_sConversationState.inputSubmittedCondition.notify_all();
     if (m_thread.joinable())
         m_thread.join();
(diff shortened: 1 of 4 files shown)


MaiteALC:
### How I tested

I'm not familiar with the C++ ecosystem and its build system, so I asked Google Gemini for the build commands, and they tell me to run:

1. `cmake --no-warn-unused-cli -DCMAKE_BUILD_TYPE:STRING=Release -S . -B ./build`

2. `cmake --build ./build --config Release --target hyprlock -j$(nproc)`

Then, I switched my hypridle config file to use the new hyprlock I locally compiled and ran `systemctl suspend` to reproduce the issue.

---

### Results

It always works, but always lags.

The screen freezes for about 2 or 3 seconds in an intermediate state between the lock screen and the home screen (in its last state before the suspension), and then, I can type my password.

None coredumpctl logs has been produced. I think it means no errors has occurred, hyprlock only lags now.

MaiteALC:
Tell me if there is something I can do to better test your fix

g0spel:
Tested on AMD Barcelo laptop (Omarchy + CachyOS kernel, Hyprland 0.56.1-3.1). Built from 2373350 with cmake -B build -DCMAKE_BUILD_TYPE=Release, put the binary in ~/.local/bin so it shadows the distro hyprlock.
Before this fix, the exact repro was: lid close → resume → enter password → unlock works, but hyprlock stays as a zombie forever. pidof hyprlock still matches, hyprctl locked is false, no lock surface, and every subsequent lock attempt becomes a no-op because my distro's lock script guards with pidof hyprlock. I captured coredumps of two of these zombies: the main thread was stuck in std::thread::join() (futex wait on the PAM thread's TCB), and the PAM thread was sitting in pam_authenticate() → pam_unix → pam_prompt → conv waiting on a condition variable. The log always stopped at "Unlocked, exiting!" — "Reached the end, exiting" never appeared.
With this build, did 4 suspend/resume cycles with unlock after each, plus a manual lock/unlock via my lock key. All of them now log the full clean-exit chain (Unlocked, exiting! → Reached the end, exiting) and no process is left behind afterwards. The deadlock is gone.
One thing worth noting separately: in 2 of the 4 suspend cycles I saw the lock request get finished immediately ("yeeten") right when the lid close coincided with hyprlock starting. Before the fix that scenario was one of my zombie coredumps; now hyprlock logs conversation failed and exits cleanly. The screen still locked and unlocked fine in those rounds (Hyprland still held the session lock), so functionally it worked — the "auth failure" there is just the terminateRequested/isFadingOutOrTerminating check in conv() doing its job on a rejected lock, not a regression from this PR. The yeeten itself looks like a separate race on the Hyprland side (suspend broadcast vs. lock request timing), but I didn't dig further into it.
So: the deadlock fix is solid on my end — no more zombies, no leftover processes, no regressions in normal lock/unlock. Thanks for the quick fix.

PR #1066, opened 2026-09-01
fingerprint: retry claim/verify after resume instead of dying silently

ZbyszekK:
Similar fix as https://github.com/hyprwm/hyprlock/pull/1049 but more complex failing scenario.

Fingerprint session goes stale after suspend if the reader re-enumerates (e.g. USB power-cycles on longer sleeps). VerifyStart on the old proxy fails silently and fingerprint auth just stays dead until a password attempt forces a reclaim.

Fix: on wake, retry Claim/VerifyStart with backoff (500ms, 10 tries max), dropping the stale device proxy on failure so a fresh one gets fetched.

Not doing a forced VerifyStop+Release on every wake (when the reader didn't lose power a.k.a short suspend), since it hits a still-live verify session. Only tears down on actual failure.

Tested on Synaptics Prometheus (06cb:00bd), short + long suspend, both recover without password fallback.

### DIFF
--- a/src/auth/Fingerprint.cpp
+++ b/src/auth/Fingerprint.cpp
@@ -71,8 +71,8 @@ void CFingerprint::init() {
     g_dbus->m_freedesktopLogin1->uponSignal("PrepareForSleep").onInterface(LOGIN_MANAGER).call([this](bool start) {
         Log::logger->log(Log::INFO, "fprint: PrepareForSleep (start: {})", start);
         m_sDBUSState.sleeping = start;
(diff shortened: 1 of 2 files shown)


ZbyszekK:
Closing as this MAX_VERIFY_RETRIES and RETRY_INTERVAL are dirty and shall not become part of the main code base. Hovewer, tha patch itself is fixing the root cause in my configuration of fingerprint device changing its device number on longer suspend.

sudosteve:
I'm confused. It fixes your issue but your closing the PR?

PR #1069, opened 2026-09-02
pam: delay auth init until grace period expires

deivi98:
During a grace period, hyprlock allows unlocking via mouse movement or a keypress without completing PAM authentication. However, PAM (and therefore any module in the stack, e.g. pam_u2f) was being initialised immediately on startup regardless of grace, causing FIDO2 security keys to light up and wait for a touch that is never needed when the user unlocks during grace.

Fix this by waiting at the start of CPam::init()'s thread until the grace period wall-clock time has elapsed before calling into pam_authenticate(). The wait uses the existing inputSubmittedCondition condvar so that terminate() wakes the thread immediately if the session is unlocked (via grace, SIGUSR1, etc.) before the grace period ends, avoiding a hang on exit.

### DIFF
--- a/src/auth/Pam.cpp
+++ b/src/auth/Pam.cpp
@@ -96,6 +96,20 @@ CPam::~CPam() {
 
 void CPam::init() {
     m_thread = std::thread([this]() {
+        // If a grace period is active, delay PAM initialisation until it expires.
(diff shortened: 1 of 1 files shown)


deivi98:
Tested locally and it works, even with system suspend mid grace period

